PR #7, opened 2024-10-28
ci: add check to enforce presence of license identifier

ingomueller-net:


ingomueller-net:
The Github action doesn't work out of the box but I submitted enarx/spdx#24 and enarx/spdx#25 that should make it work.

jpienaar (approved):
Approved, but assuming we may have to delay until those other commits are accepted?

ingomueller-net:
Yeah, exactly.

jpienaar:
I see it worked and flagged some :)

PR #8, opened 2024-10-28
Add CI check for conventional commit message.

ingomueller-net:
This PR adds a check to CI that ensures that the commit message follows the the form of "conventional commits." The file is copied from the main `substrait` repository with the small modification of fixing the version of the `npm` package that is installed.


jpienaar (approved):
"logic that posts comment upon failure" - that seems useful though. How do folks see it without that? (in the bottom CI section?)

ingomueller-net:
> "logic that posts comment upon failure" - that seems useful though. How do folks see it without that? (in the bottom CI section?)

Yes, like with all other CI targets. Having the CI target post a comment requires setting up an API token, passing that into the CI target, and giving it permissions etc., which I didn't want to read up on...

PR #13, opened 2024-10-28
docs: fix and extend build instructions

ingomueller-net:
This commit fixes several problems with the build instructions in the README and extends them. In particular, the following changes are made:

* List global prerequisites (like git, CMake, C++ compiler, etc.)
* Fix base path and define it as a first step
* Explicitly describe what paths to add to the `activate` script, including `PATH`, and adapt following commands to that
* Mention possible problems with using `lld`
* Remove the quirk with `-DABSL_ENABLE_INSTALL=ON` -- this doesn't seem to be necessary anymore
* Remove `lit` from the Python requirements because LLVM provides it in the `PATH`
* Adapt the `lit` examples to the new repository structure

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,6 @@ that don't have `lld`.
 ```bash
 cmake \
   -DPython3_EXECUTABLE=$(which python) \
-  -DABSL_ENABLE_INSTALL=ON \
   -DBUILD_SHARED_LIBS=ON \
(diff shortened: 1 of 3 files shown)


mbrobbel (approved):

ingomueller-net:
Thanks @mbrobbel for the review! (I still have to wait for a review from someone with write access...)

jpienaar (approved):

PR #17, opened 2024-11-20
feat: add a custom assembly with explicit return type for `union distinct` operation

dshaaban01:
Please only review top commit, bottom commit in [PR #16](https://github.com/substrait-io/substrait-mlir-contrib/pull/16).

Added custom assembly with explicit return type for Union Operation and adapted the tests.

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
@@ -395,9 +395,8 @@ def Substrait_UnionDistinct_Op : Substrait_RelOp<"union_distinct"> {
 
   let results = (outs Substrait_Relation:$result);
   let assemblyFormat = [{
-    $left `u_d` $right attr-dict `:` type($left) `u_d` type($right) `->` type($result)
+    $left `u` $right attr-dict `:` type($left) `u` type($right) `->` type($result)
   }];
-
 }
 
 def Substrait_EmitOp : Substrait_RelOp<"emit", [
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
@@ -394,6 +394,10 @@ def Substrait_UnionDistinct_Op : Substrait_RelOp<"union_distinct"> {
   );
 
   let results = (outs Substrait_Relation:$result);
+  let assemblyFormat = [{
+    $left `u_d` $right attr-dict `:` type($left) `u_d` type($right) `->` type($result)
+  }];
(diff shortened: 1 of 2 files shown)


test/Dialect/Substrait/union-distinct.mlir
```diff
@@ -5,15 +5,16 @@
...
+// CHECK-SAME:        : tuple<si32> u_d tuple<si32> -> tuple<si32>
 // CHECK-NEXT:      yield %[[V2]] : tuple<si32>
 
+
 substrait.plan version 0 : 42 : 1 {
   relation {
     %0 = named_table @t1 as ["a"] : tuple<si32>
     %1 = named_table @t2 as ["b"] : tuple<si32>
-    %2 = "substrait.union_distinct"(%0 , %1) : (tuple<si32>,tuple<si32>) -> tuple<si32>
+    %2 = union_distinct %0 u_d %1 : tuple<si32> u_d tuple<si32>  -> tuple<si32>
```
ingomueller-net, line 17: Nit: remove stray space.
```suggestion
    %2 = union_distinct %0 u_d %1 : tuple<si32> u_d tuple<si32> -> tuple<si32>
```

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -394,6 +394,9 @@ def Substrait_UnionDistinct_Op : Substrait_RelOp<"union_distinct"> {
   );
 
   let results = (outs Substrait_Relation:$result);
+  let assemblyFormat = [{
+    $left `u_d` $right attr-dict `:` type($left) `u_d` type($right) `->` type($result)
```
ingomueller-net, line 398: I am not a huge fan of the `u_d` "operator" here, TBH. I see how this is an evolution of the `x` in `CrossOp`, which makes sense. But currently the op name already produces the information "union" and "distinct" so the "operator" only needs to separate left from right input. `u` or `,` would be sufficient to do that. I also didn't immediate understand what `u_d` stands for.

TL;DR: For this PR, I suggest to use `u` or `,`. In some subsequent PR, we have to think about how to represent the various variants anyways, so the choice isn't hugely important.

ingomueller-net (approved):
Yep, nice, good progress, LGTM! Again, only very minor comments.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
dshaaban01: Okay makes sense. I'll go with "u".

dshaaban01:
@ingomueller-net I have successfully rebased with no casualties. ;)

PR #21, opened 2024-11-25
fix: add `substrait-translate` as dependency to `check-substrait-mlir`

dshaaban01:
`substrait-translate` not built when running `ninja check-substrait-mlir` so I added it to dependency list.


ingomueller-net (approved):
Good catch, thanks!

PR #22, opened 2024-11-27
ci: run pyright as part of CI and fix issues it found

ingomueller-net:
This PR runs `pyright` as part of CI and fixes some problems it found:

* In the DuckDB test, we converted the `str` Substrait plan to `bytes`
  even though the parameter expects a `str`, so the commit removes the
  conversion.
* In the test for converting MLIR ops to plans, we use some op-specific
  members on an object whose most narrow known type is `MlirOperation`,
  so the commit adds a `cast` to make the `typing` information more
  precise.

I have also tried to set up `pytest` but couldn't make it find all paths.

### DIFF
--- a/test/python/dialects/substrait/e2e_duckdb.py
+++ b/test/python/dialects/substrait/e2e_duckdb.py
@@ -49,6 +49,7 @@ def testNamedTable():
   pb_plan = ss.to_binpb(plan.operation).encode()
 
   # Execute in duckdb and print result.
+  # Ignore type because DuckDB's `from_substrait` has wrong type annotation.
   query_result = con.from_substrait(proto=pb_plan)  # type: ignore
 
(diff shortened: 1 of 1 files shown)


ingomueller-net:
(Finally working as intended now and ready for review.)

test/python/dialects/substrait/e2e_duckdb.py
```diff
@@ -49,7 +49,7 @@ def testNamedTable():
   pb_plan = ss.to_binpb(plan.operation).encode()
 
   # Execute in duckdb and print result.
-  query_result = con.from_substrait(proto=pb_plan)
+  query_result = con.from_substrait(proto=pb_plan)  # type: ignore
```
jpienaar, line 52: Why ignore?

jpienaar (approved):

PR #23, opened 2024-11-27
feat: implement `union_distinct` export function

dshaaban01:
Implemented `union_distinct` export function and added test.


ingomueller-net (approved):
LGTM!

(I had not seen this PR before; I think I hadn't been marked as reviewer and #24 didn't mention it. But I did see it now so all good 😄)

PR #24, opened 2024-11-27
feat: implement import function for `union_distinct`

dshaaban01:
Implemented import function for `union_distinct` and a test case for it. Added round trip capability in export test. 

Please only review latest commit. Other commit in PR_23 (https://github.com/substrait-io/substrait-mlir-contrib/pull/23)

### DIFF
--- a/lib/Target/SubstraitPB/Export.cpp
+++ b/lib/Target/SubstraitPB/Export.cpp
@@ -45,6 +45,7 @@ public:
 
   DECLARE_EXPORT_FUNC(CallOp, Expression)
   DECLARE_EXPORT_FUNC(CrossOp, Rel)
+  DECLARE_EXPORT_FUNC(UnionDistinctOp, Rel)
   DECLARE_EXPORT_FUNC(EmitOp, Rel)
   DECLARE_EXPORT_FUNC(ExpressionOpInterface, Expression)
   DECLARE_EXPORT_FUNC(FieldReferenceOp, Expression)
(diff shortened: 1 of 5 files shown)


lib/Target/SubstraitPB/Export.cpp
```diff
@@ -223,6 +224,49 @@ FailureOr<std::unique_ptr<Rel>> SubstraitExporter::exportOperation(CrossOp op) {
   return rel;
 }
 
+FailureOr<std::unique_ptr<Rel>>
+SubstraitExporter::exportOperation(UnionDistinctOp op) {
```
ingomueller-net, line 228: I am trying to keep the order of the functions, the forward declarations/macros, and template parameter lists ordered. Could you move this and the other additions to the right place?

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -140,6 +141,32 @@ static mlir::FailureOr<CrossOp> importCrossRel(ImplicitLocOpBuilder builder,
   return builder.create<CrossOp>(leftVal, rightVal);
 }
 
+static mlir::FailureOr<UnionDistinctOp>
+importSetRel(ImplicitLocOpBuilder builder, const Rel &message) {
+  const SetRel &setRel = message.set();
+
+  assert(setRel.inputs_size() == 2 && "expected two inputs");
```
ingomueller-net, line 148: I think a failure is more appropriate, here and below. An `assert` is something that shouldn't happen, in particular, not due to user-provided input, and is mainly there for documentation for readers of the code. (Note that assertions are not tested if the project is compiled in release mode.) Here, the problem comes from the structure of a user-provided file and the user may want to know about that problem.

BTW, since we'll remove that limitation soon, you don't need to write a test case for that error message.

ingomueller-net (approved):
Very nice, LGTM!

lib/Target/SubstraitPB/Export.cpp
dshaaban01: I will apply this feedback in the export function PR(https://github.com/substrait-io/substrait-mlir-contrib/pull/23).

ingomueller-net:
@dshaaban01: this needs rebasing/merging before we can proceed.

dshaaban01:
> @dshaaban01: this needs rebasing/merging before we can proceed.

Done.

PR #25, opened 2024-11-27
chore: fix to make already merged `union_distinct` related functions be in alphabetical order

dshaaban01:
Following feedback from the previous PR, I put the union_distinct op type inference function and the union_distinct op declaration in their correct alphabetical locations.


ingomueller-net (approved):
LGTM! Great, thanks for taking care of this.

PR #27, opened 2024-11-29
feat: transform `union_distinct` op into `set` operation that supports all enumerations

dshaaban01:
- Changed `UnionDistinctOp` into `SetOp` that contains enum's for all the set relations. Adjusted valid and invalid test cases to include all the enumerations.
- Adjusted return type inference function to handle all enum's. 
- Adjusted export function to handle other enum's + adjusted test case to handle all enumerations.
- Adjusted import function to handle other enum's + adjusted test case to handle all enumerations.

Note: This implementation only works for two inputs. TODO: extend to support `repeatable` inputs.


dshaaban01:
@ingomueller-net 

Enums ready!

include/substrait-mlir/Dialect/Substrait/IR/CMakeLists.txt
```diff
@@ -7,8 +7,16 @@ mlir_tablegen(SubstraitOpInterfaces.cpp.inc -gen-op-interface-defs)
 mlir_tablegen(SubstraitTypeInterfaces.h.inc -gen-type-interface-decls)
 mlir_tablegen(SubstraitTypeInterfaces.cpp.inc -gen-type-interface-defs)
 add_public_tablegen_target(MLIRSubstraitInterfacesIncGen)
-add_dependencies(MLIRSubstraitDialect MLIRSubstraitInterfacesIncGen)
```
ingomueller-net, line 10: Kudos for making this work! I hadn't had this on the radar.

On detail, though: the names of the generated files don't follow the conventions yet.
And an even more minor nit: Let's move this up for alphabetical order and add an independent `add_depencencies`:

```cmake
set(LLVM_TARGET_DEFINITIONS SubstraitOps.td)
mlir_tablegen(SubstraitEnums.h.inc -gen-enum-decls)
mlir_tablegen(SubstraitEnums.cpp.inc -gen-enum-defs)
add_public_tablegen_target(MLIRSubstraitEnumsIncGen)
add_dependencies(MLIRSubstraitDialect MLIRSubstraitEnumsIncGen)
```

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -18,6 +18,7 @@ include "mlir/IR/BuiltinAttributes.td"
 include "mlir/IR/CommonAttrConstraints.td"
 include "mlir/IR/OpBase.td"
 include "mlir/IR/SymbolInterfaces.td"
+include "mlir/IR/EnumAttr.td"
```
ingomueller-net, line 21: Nit: alphabetical order.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -525,24 +526,35 @@ def Substrait_ProjectOp : Substrait_RelOp<"project", [
   }];
 }
 
-def Substrait_UnionDistinctOp : Substrait_RelOp<"union_distinct", [
+def SetOpKind : I32EnumAttr<"SetOpKind", "The enum values correspond to those in the SetRel.SetOp message.",[I32EnumAttrCase<"unspecified", 0>,
```
ingomueller-net, line 529: Nit: break into more lines (of ~80 characters, I think?).

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -717,8 +717,26 @@ SubstraitExporter::exportOperation(ProjectOp op) {
...
+    return ::substrait::proto::SetRel::SET_OP_MINUS_MULTISET;
+  case SetOpKind::intersection_primary:
+    return ::substrait::proto::SetRel::SET_OP_INTERSECTION_PRIMARY;
+  case SetOpKind::intersection_multiset:
+    return ::substrait::proto::SetRel::SET_OP_INTERSECTION_MULTISET;
+  case SetOpKind::union_distinct:
+    return ::substrait::proto::SetRel::SET_OP_UNION_DISTINCT;
+  case SetOpKind::union_all:
+    return ::substrait::proto::SetRel::SET_OP_UNION_ALL;
+  }
```
ingomueller-net, line 736: We should try to make this work with a `static_cast`. Since the respective enum values have the same integer correspondance, that should work...

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -155,16 +176,18 @@ importSetRel(ImplicitLocOpBuilder builder, const Rel &message) {
...
-  if (setRel.op() != SetRel::SET_OP_UNION_DISTINCT)
-    return mlir::emitError(
-        builder.getLoc(),
-        "only 'union distinct' set operation supported for import");
+  std::optional<SetOpKind> kind = getSetOpKind(setRel.op());
+
+  // Check for unsupported set operations.
+  if (!kind)
+    return mlir::emitError(builder.getLoc(),
+                           "Set operation not supported for import");
```
ingomueller-net, line 184: It is probably wise to make this a compiler failure rather than an assertion because it could be that the protobuf definitions get extended (and maybe it's even possible to have invalid values in protobuf). But I would change the message to something like "unexpected 'operation' found". If possible, it'd be great to specify the value as well.

test/Dialect/Substrait/set-invalid.mlir
```diff
@@ -0,0 +1,90 @@
...
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a"] : tuple<si32>
+    %1 = named_table @t2 as ["b"] : tuple<si1>
+    // expected-error@+2 {{'substrait.set' op failed to infer returned types}}
+    // expected-error@+1 {{left and right inputs must have the same field types}}
+    %2 = set union_all, %0, %1 : tuple<si32>, tuple<si1> -> tuple<si32>
+    yield %2 : tuple<si32>
+  }
```
ingomueller-net, line 89: I think I would remove those: the error message does not depend on the `operation` and having several tests that that test the same thing are just a (small) burden.

ingomueller-net (approved):
Really nice, LGTM!

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
dshaaban01: The issue is that the description, "The enum values...", is a single parameter within the constructor for the class I32EnumAttr<string name, string summary, list<I32EnumAttrCase> cases>. I think it cannot be split into multiple strings without breaking the integrity of the function call. @ingomueller-net

ingomueller-net: You have several options: (1) break the line before the string and (2) use a multline string (`[{multiline string}]`).

dshaaban01: My bad! I tried with """ """ and it did not work and I forgot about the [{ syntax! Thanks!

lib/Target/SubstraitPB/Import.cpp
dshaaban01: My impression was that returning the result of emitError would return failure?

ingomueller-net: Yes, yes, when I said "it is probably wise", I was confirming that you took the right decision. Sorry about the confusion.

include/substrait-mlir/Dialect/Substrait/IR/CMakeLists.txt
```diff
@@ -1,20 +1,21 @@
 add_mlir_dialect(SubstraitOps substrait)
 add_dependencies(MLIRSubstraitDialect MLIRSubstraitOpsIncGen)
 
+# Add SetOpKind attribute
+set(LLVM_TARGET_DEFINITIONS SubstraitOps.td)
+mlir_tablegen(SubstraitEnums.td.inc -gen-enum-decls)
```
ingomueller-net, line 6:
```suggestion
mlir_tablegen(SubstraitEnums.h.inc -gen-enum-decls)
```

ingomueller-net: Can we call the file like this and fix the name in the `#include` directive instead?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -526,7 +526,8 @@ def Substrait_ProjectOp : Substrait_RelOp<"project", [
   }];
 }
 
-def SetOpKind : I32EnumAttr<"SetOpKind", "The enum values correspond to those in the SetRel.SetOp message.",[I32EnumAttrCase<"unspecified", 0>,
+def SetOpKind : I32EnumAttr<"SetOpKind", [{The enum values correspond to those 
+in the SetRel.SetOp message.}],[I32EnumAttrCase<"unspecified", 0>,
```
ingomueller-net, line 530: I'd prefer something like this:
```suggestion
def SetOpKind : I32EnumAttr<"SetOpKind",
  "The enum values correspond to those in the SetRel.SetOp message.", [
    I32EnumAttrCase<"unspecified", 0>,
    # Remaining cases indented by four spaces
```

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -769,7 +750,7 @@ FailureOr<std::unique_ptr<Rel>> SubstraitExporter::exportOperation(SetOp op) {
   setRel->set_allocated_common(relCommon.release());
   setRel->add_inputs()->CopyFrom(*rightRel->get());
   setRel->add_inputs()->CopyFrom(*leftRel->get());
-  setRel->set_op(getSetOp(op.getKind()));
+  setRel->set_op(static_cast<::substrait::proto::SetRel::SetOp>(op.getKind()));
```
ingomueller-net, line 753: Nice :)

ingomueller-net (approved):
LGTM modulo even more minor things!

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
ingomueller-net: I just looked at the failures in CI. It could be that multiline strings aren't allowed in that case (but my diff also happens to solve that). I *think* you can also do something like that (in which case there is no line break in the defined string):

```tablegen
"first line"
"second line"
```

include/substrait-mlir/Dialect/Substrait/IR/CMakeLists.txt
dshaaban01: Ah now I understand, thanks!

dshaaban01:
Okay fixed all new comments :) @ingomueller-net

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -526,8 +526,10 @@ def Substrait_ProjectOp : Substrait_RelOp<"project", [
   }];
 }
 
-def SetOpKind : I32EnumAttr<"SetOpKind", [{The enum values correspond to those 
-in the SetRel.SetOp message.}],[I32EnumAttrCase<"unspecified", 0>,
+def SetOpKind : I32EnumAttr<"SetOpKind",
+  "The enum values correspond to those in the SetRel.SetOp message.", [
+    I32EnumAttrCase<"unspecified", 0>,
+  // Remaining cases indented by four spaces
```
ingomueller-net, line 532: Haha, sorry, that wasn't clear enough. I meant that to be an instruction to you: please indent the remaining cases by four spaces rather than two such that they are aligned with the `unspecified` case.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -35,7 +35,6 @@ void SubstraitDialect::initialize() {
 }
 
 #include "substrait-mlir/Dialect/Substrait/IR/Substrait.cpp.inc"
-
```
ingomueller-net: Ultra nit: What do the other such `include` directives do? Do they have surrounding blank lines? If yes, I'd leave that one as well.

dshaaban01:
I'm not sure why this is failing... It builds locally on my laptop.

include/substrait-mlir/Dialect/Substrait/IR/Substrait.h
```diff
@@ -14,6 +14,7 @@
 #include "mlir/IR/OpImplementation.h"             // IWYU: keep
 #include "mlir/IR/SymbolTable.h"                  // IWYU: keep
 #include "mlir/Interfaces/InferTypeOpInterface.h" // IWYU: keep
+#include "substrait-mlir/Dialect/Substrait/IR/SubstraitEnums.h.inc"
```
ingomueller-net, line 17: Sorry, only seeing this now: add a new line before this one, please, and add ` // IWYU: export` at the end. (The lines before are `include`s used by the generated code and the `include`s of the generated files "export" that generated file, so they belong to two different blocks. The `IWYU` pragma is for a tool, `include-what-you-use`, that otherwise wouldn't understand why we include that header since we are not using it. We aren't running that tool currently, so it doesn't matter to much, TBH.)

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -141,8 +141,29 @@ static mlir::FailureOr<CrossOp> importCrossRel(ImplicitLocOpBuilder builder,
...
+  case ::substrait::proto::SetRel::SET_OP_INTERSECTION_MULTISET:
+    return SetOpKind::intersection_multiset;
+  case ::substrait::proto::SetRel::SET_OP_UNION_DISTINCT:
+    return SetOpKind::union_distinct;
+  case ::substrait::proto::SetRel::SET_OP_UNION_ALL:
+    return SetOpKind::union_all;
+  default:
+    return std::nullopt;
+  }
+}
```
ingomueller-net, line 163: We don't call this function in the end, right? Then we should remove it, no?

ingomueller-net (approved):
LGTM! (Still a few even more minor points 😉 )

ingomueller-net:
> I'm not sure why this is failing... It builds locally on my laptop.

This can happen if you change the name of a generated file: then the file with the old name still exists in your build folder, so references to the file with the old name still work, but that file won't be generated in CI. You can reproduce the CI behavior by removing a subfolder (!) of the build folder that contains that file and test again.

lib/Target/SubstraitPB/Import.cpp
dshaaban01: We use this in the import function. The one we discussed/I removed was in the export function.

dshaaban01:
> > I'm not sure why this is failing... It builds locally on my laptop.
> 
> This can happen if you change the name of a generated file: then the file with the old name still exists in your build folder, so references to the file with the old name still work, but that file won't be generated in CI. You can reproduce the CI behavior by removing a subfolder (!) of the build folder that contains that file and test again.

Okay got it, good to know - thank you!

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
dshaaban01: Oopsie sorry about that 😆

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: Ah, OK, now I was too quick... I think the static cast should work the other way as well. Can you try?

dshaaban01: On it.

ingomueller-net (approved):
LGTM! No nits left 😉

dshaaban01:
> LGTM! No nits left 😉

Can we merge 🤓? Want to start join off of main so I can leverage enum header files :)

ingomueller-net:
Merrrrrrged 🎉

PR #28, opened 2024-12-02
feat: introduce `fetch` operation

dshaaban01:
Introduces the `fetch` operation, including return type inference, import/export functions and respective tests.

@ingomueller-net


ingomueller-net:
> Why has this been building for an hour? 🫢

It occasionally happens (for reasons unknown to me) that a CI job doesn't have a build cache. Then it makes a scratch build, witch takes ~1:45h...

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -155,6 +155,22 @@ EmitOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+                          OpaqueProperties properties, RegionRange regions,
+                          llvm::SmallVectorImpl<Type> &inferredReturnTypes) {
+  Value input = operands[0];
+
+  TypeRange leftFieldTypes = cast<TupleType>(input.getType()).getTypes();
+
+  auto resultType = TupleType::get(context, leftFieldTypes);
+
+  inferredReturnTypes = SmallVector<Type>{resultType};
+
```
ingomueller-net, line 170: Nit: there are too many blank lines here. I'd say you don't need any of them; it'll still be just five and all pretty straightforward...

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -155,6 +155,22 @@ EmitOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+                          ValueRange operands, DictionaryAttr attributes,
+                          OpaqueProperties properties, RegionRange regions,
+                          llvm::SmallVectorImpl<Type> &inferredReturnTypes) {
+  Value input = operands[0];
+
+  TypeRange leftFieldTypes = cast<TupleType>(input.getType()).getTypes();
+
+  auto resultType = TupleType::get(context, leftFieldTypes);
+
+  inferredReturnTypes = SmallVector<Type>{resultType};
```
ingomueller-net, line 169: The way how this interface is intended is to fill the `SmallVector` that is passed in, so you'd `inferredReturnTypes.push_back(resultType)`. In this case, the function is probably always being called with an empty vector but I think it's still better to stick to the convention.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -155,6 +155,22 @@ EmitOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
   return success();
 }
 
+LogicalResult
+FetchOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
+                          ValueRange operands, DictionaryAttr attributes,
+                          OpaqueProperties properties, RegionRange regions,
+                          llvm::SmallVectorImpl<Type> &inferredReturnTypes) {
+  Value input = operands[0];
```
ingomueller-net, line 163: Minor nit: it is *maybe* more readable to say `Type inputType = ...` here and remove the `getType` from the next line...?

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -273,6 +274,19 @@ importLiteral(ImplicitLocOpBuilder builder,
   }
 }
 
+static mlir::FailureOr<FetchOp> importFetchRel(ImplicitLocOpBuilder builder,
+                                               const Rel &message) {
+  // Import input.
+  const Rel &inputRel = fetchRel.input();
+
```
ingomueller-net, line 283: Nit: remove this blank line and the one between the next two statements.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -415,6 +415,25 @@ def Substrait_EmitOp : Substrait_RelOp<"emit", [
...
+  ]> {
+  let summary = "Fetch operation";
+  let description = [{
+    Represents a `FetchOp` message together with the `RelCommon` and input
+    `Rel` message it contains. 
+  }];
+  let arguments = (ins
+    Substrait_Relation:$input,
+    I64Attr: $offset,
+    I64Attr: $count
```
ingomueller-net, line 429: There are actually constraints for what values these numbers can take. From the spec:

| Property | Description | Required |
| ----- | ------------ | ----------- |
| Offset | A non-negative integer. Declares the offset for retrieval of records. | Optional, defaults to 0. |
| Count | A non-negative integer or -1. Declares the number of records that should be returned. -1 signals that ALL records should be returned. | Required |

We don't need to verify this in this PR but if we don't, then we should add a `TODO`.

The way to do this is with [attribute constraints](https://mlir.llvm.org/docs/DefiningDialects/Operations/#single-entity-constraint). `offset` probably just needs [`IntNonNegative`](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/CommonAttrConstraints.td#L780). For `count`, you'd also need to allow `-1`, which you could express as "(larger than -2) and (not equal to -1)", which I believe can be composed with existing tablegen constraints.

ingomueller-net: It may also be worth it to make `offset` a default-valued attribute. This would make tablegen create C++ builders that omit that argument and set it to the default value (`0`).

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -415,6 +415,25 @@ def Substrait_EmitOp : Substrait_RelOp<"emit", [
...
+  );
+  let results = (outs Substrait_Relation:$result);
+  let assemblyFormat = [{
+    $input `,` $offset `,` $count attr-dict `:` type($input) `->` type($result)
```
ingomueller-net, line 433: Bonus: input and output type are always the same. This means that we can omit the type of one of the two in the assembly format if we specify the equality of the types using one of the constraints mentioned [here](https://mlir.llvm.org/docs/DefiningDialects/Operations/#type-inference). (That would also allow us to remove the `inferReturnType` function, I believe.)

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -155,6 +155,22 @@ EmitOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
   return success();
 }
 
+LogicalResult
+FetchOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
```
ingomueller-net, line 159: It may be possible to remove this function and do type inference declaratively. See comment in `SubstraitOps.td`.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
ingomueller-net: I think we can be a bit more creative here. For example `fetch 5 offset 10 from %input : tuple<si32>`. WDYT?

Super extra bonus: it'd be great if we can skip the part of `offset` part if it is 0 and/or print the `count` as `all` if the value is `-1`. I am pretty sure the latter requires a custom printer/parser directive, so maybe skip that in this PR.

ingomueller-net (approved):
Yesss! You are routined now :) LGTM!

I have a couple of comments but, as always, they are all pretty minor.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
dshaaban01: I think  "(larger than -2) and (not equal to -1)" is contradictory with wanting to allow -1? I think I could just do `ConfinedAttr<I64Attr, [IntMinValue<-1>]>`?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
dshaaban01: Will do the super extra bonus for the count part in a different PR. 😄 Managed to do the offset part in this PR.

ingomueller-net:
BTW, I've done some thinking: if you want to start extending `-substrait-emit-deduplication`, then this op is probably the best way to start. You'll get through the process of adding a pattern and associated tests but the pattern itself is probably really simple. The other ops are all of that plus a more complicated pattern...

dshaaban01:
@ingomueller-net  Should we merge or anything outstanding? 🚀

ingomueller-net:
Go for it!

dshaaban01:
I can't :( I don't have write access so I can't merge. But you go for it!! 🚀

ingomueller-net:
Oh, sorry, I forgot. Will do.

PR #29, opened 2024-12-02
feat: extend `setrel` to support variadic inputs

dshaaban01:
Extend `setrel` to support more than two inputs (now it's variadic). @ingomueller-net 

please only review [top commit](https://github.com/substrait-io/substrait-mlir-contrib/pull/29/commits/cce36198f9afdcbf8c0c8d9f049dd9f54f796c8b) (cce3619)

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
@@ -19,6 +19,7 @@ include "mlir/IR/CommonAttrConstraints.td"
 include "mlir/IR/EnumAttr.td"
 include "mlir/IR/OpBase.td"
 include "mlir/IR/SymbolInterfaces.td"
+include "mlir/IR/EnumAttr.td"
 
 class Substrait_Op<string mnemonic, list<Trait> traits = []> :
   Op<Substrait_Dialect, mnemonic, traits> {
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
@@ -548,6 +548,7 @@ def Substrait_SetOp : Substrait_RelOp<"set", [
   }];
   
   let arguments = (ins
+    // ConfinedType<Variadic<Substrait_Relation>, CPred<"($_self.size()>2)">>: $inputs,
     Variadic<Substrait_Relation>:$inputs,
     SetOpKind:$kind
   );
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
@@ -529,13 +529,12 @@ def Substrait_ProjectOp : Substrait_RelOp<"project", [
 def SetOpKind : I32EnumAttr<"SetOpKind",
   "The enum values correspond to those in the SetRel.SetOp message.", [
     I32EnumAttrCase<"unspecified", 0>,
-  // Remaining cases indented by four spaces
-  I32EnumAttrCase<"minus_primary", 1>,
-  I32EnumAttrCase<"minus_multiset", 2>,
-  I32EnumAttrCase<"intersection_primary", 3>,
-  I32EnumAttrCase<"intersection_multiset", 4>,
-  I32EnumAttrCase<"union_distinct", 5>,
-  I32EnumAttrCase<"union_all", 6>] >;
+    I32EnumAttrCase<"minus_primary", 1>,
+    I32EnumAttrCase<"minus_multiset", 2>,
+    I32EnumAttrCase<"intersection_primary", 3>,
+    I32EnumAttrCase<"intersection_multiset", 4>,
+    I32EnumAttrCase<"union_distinct", 5>,
+    I32EnumAttrCase<"union_all", 6>] >;
 
 def Substrait_SetOp : Substrait_RelOp<"set", [
     DeclareOpInterfaceMethods<InferTypeOpInterface>
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
@@ -548,7 +548,6 @@ def Substrait_SetOp : Substrait_RelOp<"set", [
   }];
   
   let arguments = (ins
-    // ConfinedType<Variadic<Substrait_Relation>, CPred<"($_self.size()>2)">>: $inputs,
     Variadic<Substrait_Relation>:$inputs,
     SetOpKind:$kind
   );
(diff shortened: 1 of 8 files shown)


include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -539,24 +539,22 @@ def SetOpKind : I32EnumAttr<"SetOpKind",
...
     `Rel`, and `SetOp` enumeration it contains. This operation performs 
     relational set operations such as intersection, union, and difference.
-    The current implementation only allows for two inputs (instead of two or more).
+    There must be at least two inputs.
   }];
   
   let arguments = (ins
-    Substrait_Relation:$left,
-    Substrait_Relation:$right,
+    Variadic<Substrait_Relation>:$inputs,
```
ingomueller-net, line 551: You do this test in `inferReturnTypes` and you may *have* to do it there (depending on the order of the verifiers and the type inference, which I don't know by heart) but I would *also* add a constraint on `inputs` here that says that there must be at least two inputs. I think there should be something in `OpBase.td` that does that. Having it here shows it in the typical first place you look up the definition of an op, so at the very least it's documentation.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -262,18 +262,20 @@ SetOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+  ValueRange inputs = operands;
 
-  if (leftFieldTypes != rightFieldTypes)
-    return ::emitError(loc.value())
-           << "left and right inputs must have the same field types";
+  if (inputs.size() < 2)
+    return ::emitError(loc.value()) << "expected at least 2 inputs";
 
-  auto resultType = TupleType::get(context, leftFieldTypes);
+  TypeRange fieldType = cast<TupleType>(inputs[0].getType()).getTypes();
```
ingomueller-net, line 270: Nit: `fieldTypes` (to reflect that it's a range).

ingomueller-net: Actually, maybe even `firstInputFieldTypes` or `referenceFieldTypes`, `refFieldTypes`, `expectedFieldTypes`?

ingomueller-net (approved):
LGTM with minor comments.

(I have been getting "out-of-date" errors while reviewing. Maybe I need to review the newest version again.)

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -262,18 +262,20 @@ SetOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
 
+  for (Value input : inputs) {
+    TypeRange inputFieldTypes = cast<TupleType>(input.getType()).getTypes();
+    if (fieldType != inputFieldTypes)
```
ingomueller-net, line 274: I hadn't realized that in the previous PR: using `!=` is exactly the right thing to do currently. However, eventually, we'll need to test for "the types are equal if we ignore nullability." Similarly, we'll need to assemble the result type such that we make any field nullable that is nullable in one of the inputs (possibly depending on the op kind). Add a `TODO` that reminds us of that.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -262,18 +262,20 @@ SetOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+      return ::emitError(loc.value())
+             << "all inputs must have the same field types";
+  }
+  auto resultType = TupleType::get(context, fieldType);
```
ingomueller-net, line 278: Nit: `TupleType::get` has non-trivial costs because it needs to look up the newly generated type in the context. (There is always only one instance of any possible type such that we can compare types by comparing the pointers those types.) Currently, there is a cheaper alternative, namely `inputs[0].getType()` (which should live in a dedicated variable where used first).

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include <google/protobuf/text_format.h>
 #include <google/protobuf/util/json_util.h>
+#include <mlir/IR/ValueRange.h>
```
ingomueller-net, line 19: I was told that MLIR and LLVM (1) always include other LLVM or MLIR headers using `""` and (2) we shouldn't include some of the most basic headers because they are always included transitively anyways (i.e., LLVM does not follow "include what you use"). That line was probably added by VS Code automatically... TL;DR: Remove that line.

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -723,35 +724,24 @@ FailureOr<std::unique_ptr<Rel>> SubstraitExporter::exportOperation(SetOp op) {
...
   setRel->set_op(static_cast<SetRel::SetOp>(op.getKind()));
 
+  // Build `inputs` message.
+  for (Value input : op.getInputs()) {
+    auto inputOp =
+        llvm::dyn_cast_if_present<RelOpInterface>(input.getDefiningOp());
+    if (!inputOp)
+      return op->emitOpError(
+          "inputs were not produced by Substrait relation op");
+    FailureOr<std::unique_ptr<Rel>> inputRel = exportOperation(inputOp);
```
ingomueller-net, line 741: We need to propagate a failure here.

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -18,6 +19,7 @@
 #include <google/protobuf/descriptor.h>
 #include <google/protobuf/text_format.h>
 #include <google/protobuf/util/json_util.h>
+#include <mlir/IR/ValueRange.h>
```
ingomueller-net, line 22: Same as above.

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -9,6 +9,7 @@
 #include "substrait-mlir/Target/SubstraitPB/Import.h"
 
 #include "ProtobufUtils.h"
+#include "google/protobuf/message_lite.h"
```
ingomueller-net, line 12: Out of curiosity: what do we need this header for?

ingomueller-net (approved):
LGTM!

(Some of my comments indeed got lost. Second time around I better found a better formulation, though 😉)

lib/Target/SubstraitPB/Import.cpp
dshaaban01: I actually didn't add this header but when I was coding and writing out various types, I think it automatically got added (this has happened in the past where some random headers I never added sneak in.) Will remove it 👍

dshaaban01:
Addressed all your comments except the one about constraining on inputs, as I did not find a constraint in OpBase.td that does that. Tried to use CPred but had weird things happen:
`ConfinedType<Variadic<Substrait_Relation>, CPred<"($_self.size()>2)">>: $inputs,`

Will have another look at it tomorrow. :)

ingomueller-net:
Hmm... I didn't find anything from my phone either. i think it can't be done as a type constraint. Maybe it works as an op constraint, where I think you can access the operand with `$input`. Don't try for too long, though: if it's not easy to do, it's probably also not easy to read, and that was the purpose...

dshaaban01:
I tried to figure out how to do it and unfortunately was not able to :( Maybe we can take a look together or I could do this in a subsequent PR? Should I add a TODO in this one?

PR #30, opened 2024-12-03
feat: use tablegen for type inference for `setop`

dshaaban01:
Implemented type inference for the `setop` using tablegen (inspo to do this from your feedback on the `fetch` PR). @ingomueller-net 

Please only review [top commit](https://github.com/substrait-io/substrait-mlir-contrib/pull/30/commits/9664db3ad72905a823bfb69c864e7651e561184d)


ingomueller-net (commented):
Execution is perfect for the PR.

I am split on whether we should do this change, though: as I commented in one of the other PRs, we'll have to change the type inference logic once we support nullability. At that point, we can use `SameOperandsAndResultType` and we almost certainly want all types to show up in the assembly. Then we'd basically revert this PR. Admittedly, the whole code base becomes a bit simpler until we get there, though. WDYT?

dshaaban01:
> Execution is perfect for the PR.
> 
> I am split on whether we should do this change, though: as I commented in one of the other PRs, we'll have to change the type inference logic once we support nullability. At that point, we can use `SameOperandsAndResultType` and we almost certainly want all types to show up in the assembly. Then we'd basically revert this PR. Admittedly, the whole code base becomes a bit simpler until we get there, though. WDYT?

I propose we push this PR now. We can always revert it when we're ready and support nullability. Keeping the codebase simpler until these changes are necessary seems like the better approach. But of course, it's your call - I think there are reasonable arguments for both approaches :)

ingomueller-net:
OK, sounds good. Let's push it then. Add a small TODO at the top of the TD definition of the op?

ingomueller-net:
Ready to rebase, I think.

dshaaban01:
> Ready to rebase, I think.

Done :)

ingomueller-net (approved):
LGTM!

PR #31, opened 2024-12-06
feat: add custom parser/printer for `fetch` count operand 

dshaaban01:
Added a custom parser and printer for the `fetch` count operand. When count is -1  it is replaced with the word "all" in the assembly language.
@ingomueller-net

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
@@ -431,7 +431,7 @@ def Substrait_FetchOp : Substrait_RelOp<"fetch", [
   );
   let results = (outs Substrait_Relation:$result);
   let assemblyFormat = [{
-    $count (`offset` $offset^)? `from` $input attr-dict `:` type($input)
+    custom < CountAsAll > ($count) (`offset` $offset^)? `from` $input attr-dict `:` type($input)
   }];
 }
 
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -44,6 +44,32 @@ void SubstraitDialect::initialize() {
 #include "substrait-mlir/Dialect/Substrait/IR/SubstraitTypeInterfaces.cpp.inc"
 
 //===----------------------------------------------------------------------===//
+// Custom Parser and Printer for Substrait
+//===----------------------------------------------------------------------===//
+
+bool parseCountAsAll(OpAsmParser &parser, IntegerAttr &count) {
+  if (!parser.parseOptionalKeyword("all")) {
+    count = parser.getBuilder().getI64IntegerAttr(-1);
+  } else {
+    int64_t result; 
+    if (!parser.parseInteger(result)) {
+      count = parser.getBuilder().getI64IntegerAttr(result);
+    } else {
+      return true;
+    } 
+  }
+  return false;
+}
+
+void printCountAsAll(OpAsmPrinter &printer, Operation* op, IntegerAttr count) {
+  if (count.getInt() == -1){
+    printer << "all";
+    return;
+  } else {
+    printer << count.getValue();
+  }
+}
+//===----------------------------------------------------------------------===//
 // Substrait operations
 //===----------------------------------------------------------------------===//
 
--- a/test/Dialect/Substrait/fetch.mlir
+++ b/test/Dialect/Substrait/fetch.mlir
@@ -19,6 +19,21 @@ substrait.plan version 0 : 42 : 1 {
 // CHECK-LABEL: substrait.plan
 // CHECK:         relation
 // CHECK:           %[[V0:.*]] = named_table
+// CHECK-NEXT:      %[[V1:.*]] = fetch all offset 3 from %[[V0]]
+// CHECK-SAME:        : tuple<si32> 
+// CHECK-NEXT:      yield %[[V1]] : tuple<si32>
+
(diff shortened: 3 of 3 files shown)


lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -43,6 +43,32 @@ void SubstraitDialect::initialize() {
...
+    count = parser.getBuilder().getI64IntegerAttr(-1);
+  } else {
+    int64_t result;
+    if (!parser.parseInteger(result)) {
+      count = parser.getBuilder().getI64IntegerAttr(result);
+    } else {
+      return true;
+    }
+  }
+  return false;
```
ingomueller-net, line 61: The control flow is a bit hard to understand here. It is often better to handle cases that return and continue without indentention/`else` branch:

```suggestion
  // `all` keyword (corresponds to `-1`).
  if (!parser.parseOptionalKeyword("all")) {
    count = parser.getBuilder().getI64IntegerAttr(-1);
    return false
  }
  
  // Normal integer.
  int64_t result;
  if (!parser.parseInteger(result)) {
    count = parser.getBuilder().getI64IntegerAttr(result);
    return false;
  }
  
  // Error.
  return true;
```

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -431,7 +431,7 @@ def Substrait_FetchOp : Substrait_RelOp<"fetch", [
   );
   let results = (outs Substrait_Relation:$result);
   let assemblyFormat = [{
-    $count (`offset` $offset^)? `from` $input attr-dict `:` type($input)
+    custom < CountAsAll > ($count) (`offset` $offset^)? `from` $input attr-dict `:` type($input)
```
ingomueller-net, line 434: Nit: no spaces.
```suggestion
    custom<CountAsAll>($count) (`offset` $offset^)? `from` $input attr-dict `:` type($input)
```
Also: does this still fit in 80 characers?

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -43,6 +43,32 @@ void SubstraitDialect::initialize() {
...
+}
+
+void printCountAsAll(OpAsmPrinter &printer, Operation *op, IntegerAttr count) {
+  if (count.getInt() == -1) {
+    printer << "all";
+    return;
+  } else {
+    printer << count.getValue();
+  }
```
ingomueller-net, line 70: Similar here: if you want to keep the `else` branch, then also `return` from it. But I think this is clearer:
```suggestion
  }
  
  // Normal integer.
  printer << count.getValue();
```

test/Dialect/Substrait/fetch.mlir
```diff
@@ -16,6 +16,21 @@ substrait.plan version 0 : 42 : 1 {
...
+// CHECK:         relation
+// CHECK:           %[[V0:.*]] = named_table
+// CHECK-NEXT:      %[[V1:.*]] = fetch all offset 3 from %[[V0]]
+// CHECK-SAME:        : tuple<si32> 
+// CHECK-NEXT:      yield %[[V1]] : tuple<si32>
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a"] : tuple<si32>
+    %1 = fetch -1 offset 3 from %0 : tuple<si32>
```
ingomueller-net, line 29: We should at least *also*  test the parsing of `all`. Was that an oversight or did you want to check the correspondence to `-1`? If so, add another test case for `all`.
```suggestion
    %1 = fetch all offset 3 from %0 : tuple<si32> 
```

ingomueller-net (approved):
Nice, you found out about custom printer/parser directives :) LGTM!

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
dshaaban01: It does not fit in 80 characters. If I split into two lines that doesn't work with assembly format right?

test/Dialect/Substrait/fetch.mlir
dshaaban01: Added an additional test case.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
ingomueller-net: Yes, it does. It already uses the multi-line string format (`[{}]`) and line breaks in the assembly format are treated the same as any other whitespace, i.e., it results in one blank space character. So you can write:

```tablegen
    custom<CountAsAll>($count) (`offset` $offset^)? `from` $input
        attr-dict `:` type($input)
```

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -431,7 +431,7 @@ def Substrait_FetchOp : Substrait_RelOp<"fetch", [
   );
   let results = (outs Substrait_Relation:$result);
   let assemblyFormat = [{
-    $count (`offset` $offset^)? `from` $input attr-dict `:` type($input)
+      custom<CountAsAll>($count) (`offset` $offset^)? `from` $input attr-dict `:` type($input)
```
ingomueller-net, line 434: Ultra nit: why two new spaces in the beginning?

ingomueller-net (approved):
LGTM!

dshaaban01:
Done @ingomueller-net. I think now ready to ship 🚀

PR #33, opened 2024-12-11
fix: documentation error in `cross op`

dshaaban01:
found a documentation error in cross op. @ingomueller-net


ingomueller-net (approved):
LGTM, good catch and thanks for fixing this!

PR #35, opened 2024-12-26
chore: remove comma from assembly language of `set` op

dshaaban01:
remove comma from assembly language of `set` op to make it cleaner.  

change from:
` %2 = set union_all, %0, %1 : tuple<si32>`
to:
` %2 = set union_all %0, %1 : tuple<si32>`

the extra comma felt unnecessary. I am also doing the `join` op assembly language in the same format as the non-comma version of this `set` op assembly language so thought I would make it the same. @ingomueller-net let me know what you think and if you agree (or not :)


ingomueller-net (approved):
Agree, this is an improvement! LGTM!

PR #39, opened 2024-12-27
chore: fix incorrect export test for `project` op

dshaaban01:
There seems to have been a mix-up: the same code used for the dialect test of the `project` operation was mistakenly included in the export test for the `project` operation. I’ve now added the correct test that accurately evaluates the export functionality of the `project` operation.

@ingomueller-net


ingomueller-net (approved):
Ha! Good catch! I must have forgotten to adapt the stuff I copied... XD

PR #41, opened 2024-12-27
feat: add `string` type

dshaaban01:
Added `string` types + `string` attribute types.

This is a stacked PR, please only review latest commit (15d2e21744b2ace928e499c5d2df8b0bc1bffcae)

@ingomueller-net

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
@@ -12,7 +12,7 @@
 include "substrait-mlir/Dialect/Substrait/IR/SubstraitDialect.td"
 include "mlir/IR/CommonTypeConstraints.td"
 include "mlir/IR/OpBase.td"
-include "mlir/IR/CommonAttrConstraints.td"
+include "mlir/IR/BuiltinAttributeInterfaces.td"
 
 // Base class for Substrait dialect types.
 class Substrait_Type<string name, string typeMnemonic, list<Trait> traits = []>
@@ -59,7 +59,7 @@ def Substrait_AtomicAttributes {
     SI64Attr, // I64
     F32Attr, // FP32
     F64Attr, // FP64
-    StrAttr, // String
+    TypedStrAttr<Substrait_String>, // String
   ];
 }
 
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
@@ -12,6 +12,7 @@
 include "substrait-mlir/Dialect/Substrait/IR/SubstraitDialect.td"
 include "mlir/IR/CommonTypeConstraints.td"
 include "mlir/IR/OpBase.td"
+include "mlir/IR/CommonAttrConstraints.td"
 
 // Base class for Substrait dialect types.
 class Substrait_Type<string name, string typeMnemonic, list<Trait> traits = []>
@@ -23,6 +24,13 @@ class Substrait_Type<string name, string typeMnemonic, list<Trait> traits = []>
 class Substrait_Attr<string name, string typeMnemonic, list<Trait> traits = []>
     : AttrDef<Substrait_Dialect, name, traits> {
   let mnemonic = typeMnemonic;
+}
+
+def Substrait_String : Substrait_Type<"String", "string"> {
+  let summary = "Substrait string type";
+  let description = [{
+    This type represents a substrait string type.
+  }];
 }
 
 /// Currently supported atomic types. These correspond directly to the types in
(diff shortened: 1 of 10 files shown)


dshaaban01:
> Note: 2 missing test files, have weird failures I will analyze later today.

The conclusion of my analysis is that the round tripping of the string attribute tests fail because it does not recognize `!substrait.string` as a type and recognizes is as `none` instead. I have therefore rebased this PR on top of the Substrait_Attr Base class PR so I can write my own custom string attr base class (currently I use MLIR built in string attr).

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -25,21 +26,39 @@ class Substrait_Attr<string name, string typeMnemonic, list<Trait> traits = []>
...
 }
 
+def Substrait_String : Substrait_Type<"String", "string"> {
+  let summary = "Substrait string type";
+  let description = [{
+    This type represents a substrait string type.
+  }];
+}
+
+def Substrait_StrAttr : Substrait_Attr<"Str", "string", [DeclareAttrInterfaceMethods<TypedAttrInterface>]> {
```
ingomueller-net, line 36: Note that there is a [`StringAttr`](https://mlir.llvm.org/docs/Dialects/Builtin/#stringattr) in the built-in dialect. I don't understand all subtleties about using it but if we don't use it, it should be a conscious decision. Did you look at it?

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -77,6 +77,14 @@ void printCountAsAll(OpAsmPrinter &printer, Operation *op, IntegerAttr count) {
   // Normal integer.
   printer << count.getValue();
 }
+
+//===----------------------------------------------------------------------===//
+// Substrait types
```
ingomueller-net, line 82:
```suggestion
// Substrait attributes
```

ingomueller-net (approved):
LGTM, modulo the question about the built-in string attribute.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
dshaaban01: Yes, I originally was using the StringAttr (built-in type) but was getting failing tests on the literals. 

"The conclusion of my analysis is that the round tripping of the string attribute tests fail because it does not recognize !substrait.string as a type and recognizes is as none instead. I have therefore rebased this PR on top of the Substrait_Attr Base class PR so I can write my own custom string attr base class (currently I use MLIR built in string attr).

My new String Attribute is called StrAttr to disambiguate from MLIR's StringAttr."

That's why I created this custom type.

ingomueller-net: I would like to understand this in more detail. Do you happen to have an old version of the code where the problem occurred? Does [this](https://github.com/llvm/llvm-project/blob/6f28b4b5e/mlir/test/IR/attribute.mlir#L420) do what you wanted to achieve? It yes, it's probably a good starting point for doing the same.

dshaaban01: Originally I was using the builtin `StrAttr` and wrote this test case.

```
// RUN: substrait-translate -substrait-to-protobuf %s \
// RUN: | FileCheck %s

// RUN: substrait-translate -substrait-to-protobuf %s \
// RUN: | substrait-translate -protobuf-to-substrait \
// RUN: | substrait-translate -substrait-to-protobuf \
// RUN: | FileCheck %s

// CHECK-LABEL: relations {
// CHECK-NEXT:    rel {
// CHECK-NEXT:      project {
// CHECK-NEXT:        common {
// CHECK-NEXT:          direct {
// CHECK-NEXT:          }
// CHECK-NEXT:        }
// CHECK-NEXT:        input {
// CHECK-NEXT:          read {
// CHECK:             expressions {
// CHECK-NEXT:          literal {
// CHECK-NEXT:            string: "hi"

  relation {
    %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.string> {
    ^bb0(%arg : tuple<si1>):
      %hi = literal "hi" : !substrait.string
      yield %hi : !substrait.string
    }
    yield %1 : tuple<si1, !substrait.string> 
  }
}
```

This test was failing and the problem I was getting was that it was inferring the return type of the literal Op as `none`, and subsequently the project op as `tuple(s1, none)` was the note/error I was getting.

ingomueller-net: I think I figured it out, see #55.

ingomueller-net (approved):

ingomueller-net:
Nice!! 🎉

PR #43, opened 2024-12-27
feat: implement `binary type`

dshaaban01:


dshaaban01:
@ingomueller-net My latest commit is failing but I don't think I am the problem. I am not sure how my changes caused this error - I pass everything locally. It could be some minor fluke I think? Could you please restart the CI?

ingomueller-net (approved):

dshaaban01:
@ingomueller-net 

All of the remaining PR's use custom attributes. I believe that they will all need to be changed just like I did in the String PR (where I created a custom String Attribute Type - see my comment [here](https://github.com/substrait-io/substrait-mlir-contrib/pull/41#discussion_r1904604253)).

Attached, see attempts at creating literal tests for binary attributes, where it is not able to parse the binary attribute value. 

Before merging any of the remaining type PR's, we should probably discuss how we want to do it. 

Screenshot 2025-01-09 at 13 10 26

Screenshot 2025-01-09 at 13 10 00

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -25,6 +25,22 @@ class Substrait_Attr<string name, string typeMnemonic, list<Trait> traits = []>
...
+    This type represents a substrait binary type.
+  }];
+}
+
+def Substrait_BinaryAttr : Substrait_Attr<"Binary", "binary"> {
+  let summary = "Substrait binary type";
+  let description = [{
+    This type represents a substrait binary attribute type.
+  }];
+  let parameters = (ins ArrayRefParameter<"uint8_t", "">:$value);
```
ingomueller-net, line 40: I hadn't realized in my first review but I think that this will be quite cumbersome in the assembly format. What do you think about making this based on a `StringAttr`? It should then work like our string type, which I found out how to do based on the built-in `StringAttr`, see #55.

dshaaban01: So, I tried the #55 approach for binary locally. 

In the `dialect/literal.mlir` test, this test passes:

// CHECK:      substrait.plan version 0 : 42 : 1 {
// CHECK-NEXT:   relation
// CHECK:         %[[V0:.*]] = named_table
// CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : tuple<si1> -> tuple<si1, !substrait.binary> {
// CHECK-NEXT:    ^[[BB0:.*]](%[[ARG0:.*]]: tuple<si1>):
**// CHECK-NEXT:      %[[V2:.*]] = literal "4,5,6,7" : !substrait.binary**
// CHECK-NEXT:      yield %[[V2]] : !substrait.binary
// CHECK-NEXT:    }
// CHECK-NEXT:    yield %[[V1]] : tuple<si1, !substrait.binary> 

  relation {
    %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.binary> {
    ^bb0(%arg : tuple<si1>):
      **%bytes = literal "\34\2c\35\2c\36\2c\37" : !substrait.binary**
      yield %bytes : !substrait.binary
    }
    yield %1 : tuple<si1, !substrait.binary> 
  }
}

In the `export/literal.mlir` and `import/literal.textpb` I get failures when I use anything (bytes, array of numbers, letter string) except for the the empty string, which passes.

Screenshot 2025-01-14 at 14 50 19

Any ideas on how to fix this?

dshaaban01:
Figured out binary!

test/Dialect/Substrait/literal.mlir
```diff
@@ -1,6 +1,30 @@
...
+// CHECK-NEXT:      yield %[[V2]] : !substrait.binary
+// CHECK-NEXT:    }
+// CHECK-NEXT:    yield %[[V1]] : tuple<si1, !substrait.binary> 
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a"] : tuple<si1>
+    %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.binary> {
+    ^bb0(%arg : tuple<si1>):
+      %bytes = literal "\34\2c\35\2c\36\2c\37" : !substrait.binary
```
ingomueller-net, line 19: This also works:

```suggestion
      %bytes = literal "4,5,6,7" : !substrait.binary
```

Is there a reason not to do it this way but to escape instead?

ingomueller-net: Nit: maybe use some other content? This might be read as an array of the bytes `4`, `5`, `6`, and `7`.

ingomueller-net (approved):

ingomueller-net (approved):

PR #45, opened 2025-01-01
feat: introduce `timestamp` and `timestamp_tz` types

dshaaban01:
Introduce both timestamp types and respective attribute types: `timestamp` and `timestamp_tz`

Note this is a stacked PR: 

`timestamp` in this commit (4b01b29b470821f585b714e33fd93066e7c155e5)

`timestamp_tz` in second commit (36761b360f73dc379b7f4f7adda7cefd281d0554)

Added tests for both timestamp types but NOT for timestamp attribute types. Waiting on https://github.com/substrait-io/substrait-mlir-contrib/pull/39 to be merged and then will add those tests to project.mlir test file (similarly to integer/float types)! @ingomueller-net

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
@@ -41,37 +41,47 @@ def Substrait_BinaryType : Substrait_Type<"Binary", "binary"> {
 }
 
 def Substrait_TimestampType : Substrait_Type<"Timestamp", "timestamp"> {
-  let summary = "Substrait timestamp (excluding timezone) type";
+  let summary = "Substrait timezone-unaware timestamp type";
   let description = [{
-    This type represents a substrait timestamp (excluding timezone) type. 
+    This type represents a substrait timezone-unaware timestamp type. 
   }];
 }
 
-def Substrait_TimestampAttr : Substrait_Attr<"Timestamp", "timestamp", 
-[DeclareAttrInterfaceMethods<TypedAttrInterface>]> {
-  let summary = "Substrait timestamp (excluding timezone) type";
+def Substrait_TimestampAttr : Substrait_Attr<"Timestamp", "timestamp",
+  [TypedAttrInterface]> {
+  let summary = "Substrait timezone-unaware timestamp type";
   let description = [{
-    This type represents a substrait timestamp (excluding timezone) attribute type.
+    This type represents a substrait timezone-unaware timestamp attribute type.
   }];  
   let parameters = (ins "int64_t":$value);
   let assemblyFormat = [{ `<` $value `` `us` `>` }];
+  let extraClassDeclaration = [{ 
+    ::mlir::Type getType() const {
+      return TimestampType::get(getContext());
+    }
+  }];
 }
 
 def Substrait_TimestampTzType : Substrait_Type<"TimestampTz", "timestamp_tz"> {
-  let summary = "Substrait timestamp (including timezone) type";
+  let summary = "Substrait timezone-aware timestamp type";
   let description = [{
-    This type represents a substrait timestamp (including timezone) type. 
+    This type represents a substrait timezone-aware timestamp type. 
   }];
 }
 
 def Substrait_TimestampTzAttr : Substrait_Attr<"TimestampTz", "timestamp_tz",
-[DeclareAttrInterfaceMethods<TypedAttrInterface>]> {
-  let summary = "Substrait timestamp (including timezone) type";
+  [TypedAttrInterface]> {
+  let summary = "Substrait timezone-aware timestamp type";
   let description = [{
-    This type represents a substrait timestamp (including timezone) attribute type.
+    This type represents a substrait timezone-aware timestamp attribute type.
   }];  
   let parameters = (ins "int64_t":$value);
   let assemblyFormat = [{ `<` $value `` `us` `>` }];
+  let extraClassDeclaration = [{ 
+    ::mlir::Type getType() const {
+      return TimestampTzType::get(getContext());
+    }
+  }];
 }
 
 /// Currently supported atomic types. These correspond directly to the types in
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
@@ -47,7 +47,8 @@ def Substrait_TimestampType : Substrait_Type<"Timestamp", "timestamp"> {
   }];
 }
 
-def Substrait_TimestampAttr : Substrait_Attr<"Timestamp", "timestamp"> {
+def Substrait_TimestampAttr : Substrait_Attr<"Timestamp", "timestamp", 
+[DeclareAttrInterfaceMethods<TypedAttrInterface>]> {
   let summary = "Substrait timestamp (excluding timezone) type";
   let description = [{
     This type represents a substrait timestamp (excluding timezone) attribute type.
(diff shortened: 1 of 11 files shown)


ingomueller-net (approved):
LGTM! (As the comment says, there are no attribute tests. Do you want to add those to this PR or to a subsequent one?)

dshaaban01:
> LGTM! (As the comment says, there are no attribute tests. Do you want to add those to this PR or to a subsequent one?)

See this [comment](https://github.com/substrait-io/substrait-mlir-contrib/pull/41#discussion_r1904604253). Happy to also do all the attribute tests in separate PR's to reduce complication - this could also be a move. Let me know.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -40,6 +40,40 @@ def Substrait_BinaryType : Substrait_Type<"Binary", "binary"> {
   }];
 }
 
+def Substrait_TimestampType : Substrait_Type<"Timestamp", "timestamp"> {
+  let summary = "Substrait timestamp (excluding timezone) type";
+  let description = [{
+    This type represents a substrait timestamp (excluding timezone) type.
```
ingomueller-net, line 46: I think the terminology from the spec is "timezone-aware" and "timezone-unaware"? If we say "including," won't it be even easier to misunderstand it to mean that the types includes the information which time zone it is, like I did this morning?

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -81,6 +81,21 @@ void printCountAsAll(OpAsmPrinter &printer, Operation *op, IntegerAttr count) {
   // Normal integer.
   printer << count.getValue();
 }
+
+//===----------------------------------------------------------------------===//
+// Substrait attributes
```
ingomueller-net, line 86: Can we move this section up, to underneath the dialect? Then it the sections (except dialect) are ordered alphabetically.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -81,6 +81,21 @@ void printCountAsAll(OpAsmPrinter &printer, Operation *op, IntegerAttr count) {
...
+//===----------------------------------------------------------------------===//
+
+/// Implement the getType method for custom type `TimestampAttr`.
+::mlir::Type TimestampAttr::getType() const {
```
ingomueller-net, line 90: You *could* also use an [`extraClassDefinition`](https://mlir.llvm.org/docs/DefiningDialects/Operations/#extra-definitions) and define this one-line function inline in the TD file.

test/Dialect/Substrait/literal.mlir
```diff
@@ -1,6 +1,32 @@
 // RUN: substrait-opt -split-input-file %s \
 // RUN: | FileCheck %s
 
+// CHECK:      substrait.plan version 0 : 42 : 1 {
+// CHECK-NEXT:   relation
+// CHECK:         %[[V0:.*]] = named_table
+// CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : tuple<si1> -> tuple<si1, !substrait.timestamp, !substrait.timestamp_tz> {
+// CHECK-NEXT:    ^[[BB0:.*]](%[[ARG0:.*]]: tuple<si1>):
+// CHECK-NEXT:      %[[V2:.*]] = literal #substrait.timestamp<10000000000us> : !substrait.timestamp
+// CHECK-NEXT:      %[[V3:.*]] = literal #substrait.timestamp_tz<10000000000us> : !substrait.timestamp_tz
```
ingomueller-net, line 10: Let's investigate a bit why the type is printed here. It is already known through the type of the attribute, so we shouldn't need it again. Parsing works without it; why is it printed? Can we get rid of it and how?

test/Target/SubstraitPB/Export/literal.mlir
```diff
@@ -9,6 +9,39 @@
...
+// CHECK-NEXT:          literal {
+// CHECK-NEXT:            timestamp_tz: 10000000000
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a"] : tuple<si1>
+    %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.timestamp, !substrait.timestamp_tz> {
+    ^bb0(%arg : tuple<si1>):
+      %timestamp = literal #substrait.timestamp<10000000000us> : !substrait.timestamp
+      %timestamp_tz = literal #substrait.timestamp_tz<10000000000us> : !substrait.timestamp_tz
```
ingomueller-net, line 36:
```suggestion
      %timestamp = literal #substrait.timestamp<10000000000us>
      %timestamp_tz = literal #substrait.timestamp_tz<10000000000us>
```

test/Dialect/Substrait/literal.mlir
ingomueller-net: Bonus: can we `CHECK` that the type isn't there once we remove it from the assembly format?

test/Target/SubstraitPB/Import/literal.textpb
```diff
@@ -10,6 +10,66 @@
 # RUN:   --split-input-file="# ""-----" --output-split-marker="// -----" \
 # RUN: | FileCheck %s
 
+# CHECK:      substrait.plan version 0 : 42 : 1 {
+# CHECK-NEXT:   relation
+# CHECK:         %[[V0:.*]] = named_table
+# CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : tuple<si1> -> tuple<si1, !substrait.timestamp, !substrait.timestamp_tz> {
+# CHECK-NEXT:    ^[[BB0:.*]](%[[ARG0:.*]]: tuple<si1>):
+# CHECK-NEXT:      %[[V2:.*]] = literal #substrait.timestamp<10000000000us> : !substrait.timestamp
+# CHECK-NEXT:      %[[V3:.*]] = literal #substrait.timestamp_tz<10000000000us> : !substrait.timestamp_tz
```
ingomueller-net, line 19: (Adapt if/when IR changed.)

ingomueller-net (approved):

test/Dialect/Substrait/literal.mlir
ingomueller-net: I think that this is related to "buildable" types. It's mentioned [here](https://mlir.llvm.org/docs/DefiningDialects/Operations/#type-inference) but I have managed to make it work.

ingomueller-net: I found out how to do it but it's somewhat involved. Will try to create a PR that lays the bases.

ingomueller-net: OK, I just pushed #59, which is still WIP. I think it will be easiest to finish up and merge this PR *with* the redundant type, then merge #59, and use it directly for all remaining types. (If we merge #59 first, we don't have anything to test the interface with, so I prefer merging this PR first.)

ingomueller-net: Actually, let's make that even more flexible: you go ahead with your type PRs with the redundant type until my PR has merge and start incorporating that from then on.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
dshaaban01: Yes agree, will fix!

lib/Dialect/Substrait/IR/Substrait.cpp
dshaaban01: Also - would you like the types to be ordered alphabetically? I thought about this yesterday... until now, I add new type and attribute definitions in order of the spec list, but not alphabetically.

test/Dialect/Substrait/literal.mlir
dshaaban01: Thank you for the PR, sounds good. Let's merge and then in a second iteration I can incorporate them. Will keep this on the TODO list.

test/Target/SubstraitPB/Import/literal.textpb
dshaaban01: Will leave this for iteration 2.

dshaaban01:
Done.

PR #46, opened 2025-01-01
feat: introduce `date` type

dshaaban01:
This PR introduces the `date` atomic type and its associated attribute types. The `date` type represents a calendar date within the range [1000-01-01..9999-12-31]. Internally, it is stored as an `int32` value representing the number of days since the epoch date (1970-01-01).

### DIFF
--- a/test/Dialect/Substrait/types.mlir
+++ b/test/Dialect/Substrait/types.mlir
@@ -3,6 +3,20 @@
 
 // CHECK-LABEL: substrait.plan
 // CHECK:         relation
+// CHECK:         %[[V0:.*]] = named_table @t1 as ["a"] : tuple<!substrait.date>
+// CHECK-NEXT:    yield %0 : tuple<!substrait.date>
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a"] : tuple<!substrait.date>
+    yield %0 : tuple<!substrait.date>
+  }
+}
+
+// -----
+
+// CHECK-LABEL: substrait.plan
+// CHECK:         relation
 // CHECK:         %[[V0:.*]] = named_table @t1 as ["a", "b"] : tuple<!substrait.timestamp, !substrait.timestamp_tz>
 // CHECK-NEXT:    yield %0 : tuple<!substrait.timestamp, !substrait.timestamp_tz>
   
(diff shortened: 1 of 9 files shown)


test/Dialect/Substrait/types.mlir
```diff
@@ -0,0 +1,14 @@
...
+// CHECK:         relation
+// CHECK:         %[[V0:.*]] = named_table @t1 as ["a"] : tuple<!substrait.date>
+// CHECK-NEXT:    yield %0 : tuple<!substrait.date>
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a"] : tuple<!substrait.date>
+    yield %0 : tuple<!substrait.date>
+  }
+}
```
ingomueller-net, line 14: Nit: final line break.

Consider [configuring](https://stackoverflow.com/a/44704969/651937) VS Code to add those line breaks automatically.

ingomueller-net (approved):
LGTM modulo minor nit.

test/Dialect/Substrait/types.mlir
dshaaban01: Haha thank you! This indeed is the smarter approach.

ingomueller-net (approved):

test/Target/SubstraitPB/Export/types.mlir
```diff
@@ -9,6 +9,62 @@
...
+// CHECK-NEXT:        named_table {
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a", "b"] : tuple<!substrait.timestamp, !substrait.timestamp_tz>
+    yield %0 : tuple<!substrait.timestamp, !substrait.timestamp_tz>
+  }
+}
+
+// -----
```
dshaaban01, line 66: @ingomueller-net I think that I accidentally duplicated the test for timestamp, timestamp_tz and that it snuck in here during the rebase :(( 

In the time PR, I basically delete the duplicated timestamp test. So thats why it looks like I'm replacing tests - it's because I deleted one. I didn't realize that it had snuck into this PR, I'm so sorry!

ingomueller-net: OK, got it. No worries, we caught and fixed it :) Thanks!

PR #48, opened 2025-01-01
feat: introduce `interval_year` and `interval_day` types

dshaaban01:
This PR introduces two new atomic interval types:

1) Interval Year to Month
Supports a range of [-10,000..10,000] years, with month-level precision. Typically stored as separate integers for years and months (e.g., 12y 3m), but only the total number of months is considered significant. For example, 1y 0m is equivalent to 0y 12m or 1001y -12000m.
2) Interval Day to Second
Supports a range of [-3,650,000..3,650,000] days, with second precision. Typically stored as separate integers for days and seconds, but only the total number of seconds is considered significant. For example, 1d 0s is equivalent to 0d 86400s.

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -75,24 +75,26 @@ def Substrait_IntervalDaySecondAttr
                                     "IntervalDaySecondType"> {
   let summary = "Substrait interval day to second type";
   let description = [{
-    This type represents a substrait interval day to second attribute type. Note that this
-    attribute does not attempt to canonicalize equivalent day-second intervals.
-  }];  
-  let parameters = (ins "int32_t":$days_value, "int32_t":$seconds_value);
-  let assemblyFormat = [{ `<` $days_value `` `d` $seconds_value `` `s` `>` }];
+    This type represents a substrait interval day to second attribute type. Note
+    that this attribute does not attempt to canonicalize equivalent day-second
+    intervals.
+  }];
+  let parameters = (ins "int32_t":$days, "int32_t":$seconds);
+  let assemblyFormat = [{ `<` $days `` `d` $seconds `` `s` `>` }];
   let genVerifyDecl = 1;
 }
 
-def Substrait_IntervalYearMonthAttr 
+def Substrait_IntervalYearMonthAttr
     : Substrait_StaticallyTypedAttr<"IntervalYearMonth", "interval_year_month",
                                     "IntervalYearMonthType"> {
   let summary = "Substrait interval year to month type";
   let description = [{
-    This type represents a substrait interval year to month attribute type. Note that this
-    attribute does not attempt to canonicalize equivalent year-month intervals.
-  }];  
-  let parameters = (ins "int32_t":$years_value, "int32_t":$months_value);
-  let assemblyFormat = [{ `<` $years_value `` `y` $months_value `` `m` `>` }];
+    This type represents a substrait interval year to month attribute type. Note
+    that this attribute does not attempt to canonicalize equivalent year-month
+    intervals.
+  }];
+  let parameters = (ins "int32_t":$years, "int32_t":$months);
+  let assemblyFormat = [{ `<` $years `` `y` $months `` `m` `>` }];
   let genVerifyDecl = 1;
 }
 
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -75,10 +75,12 @@ def Substrait_IntervalDaySecondAttr
                                     "IntervalDaySecondType"> {
   let summary = "Substrait interval day to second type";
   let description = [{
-    This type represents a substrait interval day to second attribute type.
+    This type represents a substrait interval day to second attribute type. Note that this
+    attribute does not attempt to canonicalize equivalent day-second intervals.
   }];  
   let parameters = (ins "int32_t":$days_value, "int32_t":$seconds_value);
   let assemblyFormat = [{ `<` $days_value `` `d` $seconds_value `` `s` `>` }];
+  let genVerifyDecl = 1;
 }
 
 def Substrait_IntervalYearMonthAttr 
@@ -86,10 +88,12 @@ def Substrait_IntervalYearMonthAttr
                                     "IntervalYearMonthType"> {
   let summary = "Substrait interval year to month type";
   let description = [{
-    This type represents a substrait interval year to month attribute type.
+    This type represents a substrait interval year to month attribute type. Note that this
+    attribute does not attempt to canonicalize equivalent year-month intervals.
   }];  
   let parameters = (ins "int32_t":$years_value, "int32_t":$months_value);
   let assemblyFormat = [{ `<` $years_value `` `y` $months_value `` `m` `>` }];
+  let genVerifyDecl = 1;
 }
 
 def Substrait_TimeAttr
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -70,6 +70,28 @@ def Substrait_DateAttr
   let assemblyFormat = [{ `<` $value `>` }];
 }
 
+def Substrait_IntervalDaySecondAttr
+    : Substrait_StaticallyTypedAttr<"IntervalDaySecond", "interval_day_second",
+                                    "IntervalDaySecondType"> {
+  let summary = "Substrait interval day to second type";
+  let description = [{
+    This type represents a substrait interval day to second attribute type.
+  }];  
+  let parameters = (ins "int32_t":$days_value, "int32_t":$seconds_value);
+  let assemblyFormat = [{ `<` $days_value `` `d` $seconds_value `` `s` `>` }];
+}
+
+def Substrait_IntervalYearMonthAttr 
+    : Substrait_StaticallyTypedAttr<"IntervalYearMonth", "interval_year_month",
+                                    "IntervalYearMonthType"> {
+  let summary = "Substrait interval year to month type";
+  let description = [{
+    This type represents a substrait interval year to month attribute type.
+  }];  
+  let parameters = (ins "int32_t":$years_value, "int32_t":$months_value);
+  let assemblyFormat = [{ `<` $years_value `` `y` $months_value `` `m` `>` }];
+}
+
 def Substrait_TimeAttr
     : Substrait_StaticallyTypedAttr<"Time", "time", "TimeType"> {
   let summary = "Substrait time type";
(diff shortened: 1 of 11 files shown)


test/Dialect/Substrait/types.mlir
```diff
@@ -0,0 +1,14 @@
...
+// CHECK:         relation
+// CHECK:         %[[V0:.*]] = named_table @t1 as ["a", "b"] : tuple<!substrait.interval_year, !substrait.interval_day>
+// CHECK-NEXT:    yield %0 : tuple<!substrait.interval_year, !substrait.interval_day>
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a", "b"] : tuple<!substrait.interval_year, !substrait.interval_day>
+    yield %0 : tuple<!substrait.interval_year, !substrait.interval_day>
+  }
+}
```
ingomueller-net, line 14: Nit: final line break.

ingomueller-net (approved):

ingomueller-net:
Here as well: need to rebase again.

ingomueller-net:
I think this one is next in line and needs rebasing to proceed.

dshaaban01:
Rebased and adjusted to #59. PR description also adjusted.

dshaaban01:
@ingomueller-net 

I am getting this error in CI: 

`duckdb.duckdb.HTTPException: HTTP Error: Failed to download extension "substrait" at URL "http://extensions.duckdb.org/v1.2.0/linux_amd64_gcc4/substrait.duckdb_extension.gz" (HTTP 403)
`

which is causing one test to fail:

```
Failed Tests (1):
  Substrait MLIR :: python/dialects/substrait/e2e_duckdb.py
```

I guess this is an internet connection error? Any ideas on how to fix?

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -833,6 +856,29 @@ SubstraitExporter::exportOperation(LiteralOp op) {
...
     literal->set_time(mlir::cast<TimeAttr>(value).getValue());
+  }
+  // `IntervalType`'s.
+  else if (auto timeType = dyn_cast<IntervalYearMonthType>(literalType)) {
+    auto intervalYearToMonth = std::make_unique<
+        ::substrait::proto::Expression_Literal_IntervalYearToMonth>();
+    auto intervalYear =
+        mlir::cast<IntervalYearMonthAttr>(value).getYearsValue();
+    auto intervalMonth =
+        mlir::cast<IntervalYearMonthAttr>(value).getMonthsValue();
```
ingomueller-net, line 867: Nit: use `auto` only if the type appears in the RHS of a cast or similar, see [LLVM Coding Standards](https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable). Also, my change factors out the cast, which makes it more readable and shorter. (Be sure to check if `int32_t` is the type returned by `getXValue()`.)
```suggestion
    auto intervalYearMonth = mlir::cast<IntervalYearMonthAttr>(value);
    int32_t intervalYear = intervalYearMonth.getYearsValue();
    int32_t intervalMonth = intervalYearMonth.getMonthsValue();
```

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -833,6 +856,29 @@ SubstraitExporter::exportOperation(LiteralOp op) {
...
+    intervalYearToMonth->set_years(intervalYear);
+    intervalYearToMonth->set_months(intervalMonth);
+    literal->set_allocated_interval_year_to_month(
+        intervalYearToMonth.release());
+  } else if (auto timeType = dyn_cast<IntervalDaySecondType>(literalType)) {
+    auto intervalDaytoSecond = std::make_unique<
+        ::substrait::proto::Expression_Literal_IntervalDayToSecond>();
+    auto intervalDay = mlir::cast<IntervalDaySecondAttr>(value).getDaysValue();
+    auto intervalSecond =
+        mlir::cast<IntervalDaySecondAttr>(value).getSecondsValue();
```
ingomueller-net, line 877: Same as above.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -67,6 +67,28 @@ def Substrait_DateAttr
...
+    This type represents a substrait interval day to second attribute type.
+  }];  
+  let parameters = (ins "int32_t":$days_value, "int32_t":$seconds_value);
+  let assemblyFormat = [{ `<` $days_value `` `d` $seconds_value `` `s` `>` }];
+}
+
+def Substrait_IntervalYearMonthAttr 
+    : Substrait_StaticallyTypedAttr<"IntervalYearMonth", "interval_year_month",
+                                    "IntervalYearMonthType"> {
+  let summary = "Substrait interval year to month type";
```
ingomueller-net, line 84: I think we need to document a decision here. The spec says:

> Usually stored as separate integers for years and months, but only the total number of months is significant, i.e. `1y 0m` is considered equal to `0y 12m` or `1001y -12000m`.

We *could* deduce that we only want to store the number of months, then. However, I think that our [design rationale](https://github.com/substrait-io/substrait-mlir-contrib/?tab=readme-ov-file#design-rationale) suggest we should have two different numbers:

> Every valid Substrait plan MUST be representable in the dialect.

I'd argue that two plans that differ by the representation of an `interval_year_month` are different plans (that always produce the same result), so we have to be able to express both of them with our dialect. I think that decision should be documented here.

We should, however, *eventually* implement a canonicalization pattern that converts these attributes to the representation with `month in [0,12)`.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -67,6 +67,28 @@ def Substrait_DateAttr
...
+  let description = [{
+    This type represents a substrait interval year to month attribute type.
+  }];  
+  let parameters = (ins "int32_t":$years_value, "int32_t":$months_value);
```
ingomueller-net, line 88: The [spec](https://substrait.io/types/type_classes/#simple-types) only allows "a range of [-10,000..10,000] years." If we don't add a constraint now, we should add a `TODO` to be sure we add it later.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -67,6 +67,28 @@ def Substrait_DateAttr
   let assemblyFormat = [{ `<` $value `>` }];
 }
 
+def Substrait_IntervalDaySecondAttr
```
ingomueller-net, line 70: Same two comments here as for the other interval type.

ingomueller-net (commented):
I looked at this PR again and found a few things we should think about a bit more.

dshaaban01:
Comments addressed.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -68,6 +68,23 @@ LogicalResult AdvancedExtensionAttr::verify(
   return success();
 }
 
+LogicalResult mlir::substrait::IntervalYearMonthAttr::verify(
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError, int32_t year,
+    int32_t month) {
+  if (year < -100000 || year > 100000)
+    return emitError() << "year must be in a range of [-10,000..10,000] years";
+  return success();
```
ingomueller-net, line 76: Yes, nice! `month` should be checked the same way (and be in `[120,000..120,000]`).

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -68,6 +68,23 @@ LogicalResult AdvancedExtensionAttr::verify(
...
+}
+
+LogicalResult mlir::substrait::IntervalDaySecondAttr::verify(
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError, int32_t days,
+    int32_t seconds) {
+  if (days < -3650000 || days > 3650000)
+    return emitError()
+           << "days must be in a range of [-3,650,000..3,650,000] days";
```
ingomueller-net, line 84: Same for seconds.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -70,6 +70,32 @@ def Substrait_DateAttr
...
+  let assemblyFormat = [{ `<` $days_value `` `d` $seconds_value `` `s` `>` }];
+  let genVerifyDecl = 1;
+}
+
+def Substrait_IntervalYearMonthAttr 
+    : Substrait_StaticallyTypedAttr<"IntervalYearMonth", "interval_year_month",
+                                    "IntervalYearMonthType"> {
+  let summary = "Substrait interval year to month type";
+  let description = [{
+    This type represents a substrait interval year to month attribute type. Note that this
```
ingomueller-net, line 91: Please break at 80 characters.

ingomueller-net (approved):

dshaaban01:
@ingomueller-net Did you see my comment above? We might need some more work 🫠

ingomueller-net:
Which comment? The CI failure due to the DuckDB extension? That's been fixed by #79, no?

dshaaban01:
Copy/pasted comment in response to adding a constraint for seconds

```
  if (seconds < -315360000000 || seconds > 315360000000)
    return emitError() << "seconds must be in a range of [-315,360,000,000.."
                          "315,360,000,000] seconds";
```

If I do this: 3'650'000 days * 24 hours * 60 min * 60 seconds = 315'360'000'000 seconds --> the value is larger than a int32_t max value/interval range (2,147,483,647)? Shouldn't I use an int64_t?

Screenshot 2025-02-19 at 15 21 50

This is the specification that says I should use int32_t for second and (weirdly?) microsecond precision. It's also weird because there are two different example formats in the middle and right column (middle: 1d 0s and right: 3650001d -86400s 0us)? The middle column mentions microseconds but has "s" in the format, the right column mentions seconds and microseconds - I'm confused? Could there be something wrong in the spec?

ingomueller-net:
Aha, I understand the problem. (I don't see the comment, though, but not a problem.)

As for the data type: the spec for the `Literal` submessage uses an `int32`, so I'd use that as well. If that means that we cannot get out of the valid range (or even cover it), then we don't have to test anything. I'd add a comment, though, or maybe even a `static_assert` (in the verifier, where the test would be) to be sure we think of adding a test should we ever change the type.

As for the example in the spec: the difference between the middle and the right column is that the right column has an additional `0us`, right? For consistency, I guess it would be best if the middle column had that as well. But both are just informal examples, so I'd say it's rather minor. But if you want to submit a PR that fixes the example, I am pretty sure people will appreciate 😉

dshaaban01:
Okay cool got it :)) I will add a comment in a new PR and submit a PR to fix the example :) 

Btw here's a screenshot of my comment, but it says "Pending"? So maybe by responding to your review comment, it assumed I was trying to review and because I don't have the rights to do so it doesn't display? 
Screenshot 2025-02-20 at 10 20 30

ingomueller-net:
I think what happened is that you "started a review" in the right-most tab and didn't submit the review. That review mode is handy if you have a few comments for the same PR, then they get sent as one big batch. But only once you submit it. Until then, the comments are still "pending."

PR #49, opened 2025-01-01
feat: introduce `uuid` types

dshaaban01:
This PR introduces the UUID type: a universally-unique identifier composed of 128 bits.

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
@@ -75,6 +75,15 @@ def Substrait_TimestampTzType : Substrait_Type<"TimestampTz", "timestamp_tz"> {
   }];
 }
 
+/// Currently supported atomic types. These correspond directly to the types in
+def Substrait_UUIDType : Substrait_Type<"UUID", "uuid"> {
+  let summary = "Substrait uuid type";
+  let description = [{
+    This type represents a substrait uuid (universally-unique identifier) type.
+    This is a 16-byte binary value.
+  }];
+}
+
 /// Currently supported atomic types, listed in order of substrait specification.
 /// These correspond directly to the types in
 /// https://github.com/substrait-io/substrait/blob/main/proto/substrait/type.proto.
@@ -95,6 +104,7 @@ def Substrait_AtomicTypes {
     Substrait_TimeType, // Time
     Substrait_IntervalYearMonthType, // IntervalYear
     Substrait_IntervalDaySecondType, // IntervalDay
+    Substrait_UUIDType, // UUID
   ];
 }
 
--- a/lib/Target/SubstraitPB/Export.cpp
+++ b/lib/Target/SubstraitPB/Export.cpp
@@ -362,6 +362,18 @@ SubstraitExporter::exportType(Location loc, mlir::Type mlirType) {
     return std::move(type);
   }
 
+  // Handle uuid.
+  if (mlir::isa<UUIDType>(mlirType)) {
+    // TODO(ingomueller): support other nullability modes.
+    auto uuidType = std::make_unique<proto::Type::UUID>();
+    uuidType->set_nullability(
+        Type_Nullability::Type_Nullability_NULLABILITY_REQUIRED);
+
+    auto type = std::make_unique<proto::Type>();
+    type->set_allocated_uuid(uuidType.release());
+    return std::move(type);
+  }
+
   // Handle tuple types.
   if (auto tupleType = llvm::dyn_cast<TupleType>(mlirType)) {
     auto structType = std::make_unique<proto::Type::Struct>();
@@ -949,6 +961,12 @@ SubstraitExporter::exportOperation(LiteralOp op) {
     intervalDaytoSecond->set_seconds(intervalSecond);
     literal->set_allocated_interval_day_to_second(
         intervalDaytoSecond.release());
+  } // `UUIDType`.
+  else if (auto uuidType = dyn_cast<UUIDType>(literalType)) {
+    llvm::APInt uuid = mlir::cast<UUIDAttr>(value).getValue().getValue();
+    std::string res(16, 0);
+    llvm::StoreIntToMemory(uuid, reinterpret_cast<uint8_t *>(res.data()), 16);
+    literal->set_uuid(res);
   } else
     op->emitOpError("has unsupported value");
 
--- a/test/Dialect/Substrait/literal.mlir
+++ b/test/Dialect/Substrait/literal.mlir
@@ -4,6 +4,29 @@
 // CHECK:      substrait.plan version 0 : 42 : 1 {
 // CHECK-NEXT:   relation
 // CHECK:         %[[V0:.*]] = named_table
+// CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : tuple<si1> -> tuple<si1, !substrait.uuid> {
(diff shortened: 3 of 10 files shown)


include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -25,6 +25,24 @@ class Substrait_Attr<string name, string typeMnemonic, list<Trait> traits = []>
...
+}
+
+def Substrait_UUIDAttr : Substrait_Attr<"UUID", "binary"> {
+  let summary = "Substrait uuid type";
+  let description = [{
+    This type represents a substrait uuid (universally-unique identifier) attribute type.
+    This is a 16-byte binary value.
+  }];
+  let parameters = (ins ArrayRefParameter<"uint8_t", "">:$value);
+  let assemblyFormat = [{ `<` $value `>` }];
```
ingomueller-net, line 43: Can we add a verifier/contraint to ensure it's exactly 16 bytes?

ingomueller-net, line 43: It'd be nice to (eventually) print the UUIDs in the typical format (see [Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier)). Doesn't need to be now; can just be a TODO.

If you do want to go for it, the `APInt` class from LLVM probably has a few interesting features:
* Construct from array: [constructor](https://github.com/llvm/llvm-project/blob/19c516c8d5/llvm/include/llvm/ADT/APInt.h#L148)
* Extract bits at certain positions: [`extractBits`](https://github.com/llvm/llvm-project/blob/19c516c8d/llvm/include/llvm/ADT/APInt.h#L1460).
* Convert to hex string: [`toString`](https://github.com/llvm/llvm-project/blob/19c516c8d5716c/llvm/include/llvm/ADT/APInt.h#L1663) with `radix` set to 16.

ingomueller-net (approved):

dshaaban01:
I don't know why Build and Test is failing...? Any ideas?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -75,6 +75,15 @@ def Substrait_TimestampTzType : Substrait_Type<"TimestampTz", "timestamp_tz"> {
   }];
 }
 
+/// Currently supported atomic types. These correspond directly to the types in
+def Substrait_UUIDType : Substrait_Type<"UUID", "uuid"> {
+  let summary = "Substrait uuid type";
+  let description = [{
+    This type represents a substrait uuid (universally-unique identifier) type.
+    This is a 16-byte binary value.
```
ingomueller-net, line 83: I am wondering how exactly the values are represented. I *suppose* that the string is used to represent the hexadecimal format of the 128-bit value: `c48ffa9e-64f4-44cb-ae47-152b4e60e77b`. It's important to say that -- it could also represent the value as a 16-byte string as is.

We might also think whether we don't want to use that representation. It corresponds to (what I think is) the representation of literals in protobuf plans, so it would make import and export easy.

In any case, I think it would be good to add a `def UUIDAttr` with a constrained attribute allowing only 16-byte values. (In the latter of the two cases above, that should be easy: simply a length of 16. In the former, we would probably have use some regex to make sure we only have valid values.)

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -879,6 +891,9 @@ SubstraitExporter::exportOperation(LiteralOp op) {
     intervalDaytoSecond->set_seconds(intervalSecond);
     literal->set_allocated_interval_day_to_second(
         intervalDaytoSecond.release());
+  } // `UUIDType`.
+  else if (auto uuidType = dyn_cast<UUIDType>(literalType)) {
+    literal->set_uuid(mlir::cast<StringAttr>(value).getValue().str());
```
ingomueller-net, line 896: I think that this is only valid if we interpret the attribute value as a 16-byte array as is, not as a string in hex format (see above).

test/Dialect/Substrait/literal.mlir
```diff
@@ -1,6 +1,29 @@
...
+// CHECK-NEXT:      yield %[[V2]] : !substrait.uuid
+// CHECK-NEXT:    }
+// CHECK-NEXT:    yield %[[V1]] : tuple<si1, !substrait.uuid> 
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a"] : tuple<si1>
+    %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.uuid> {
+    ^bb0(%arg : tuple<si1>):
+      %uuid = literal "c48ffa9e-64f4-44cb-ae47-152b4e60e77b" : !substrait.uuid
```
ingomueller-net, line 19: This is obviously much more readable than a 16-byte string. However, we can get the best of both world (readability and easy conversion from protobuf+more memory-efficient representation) by defining proper `UUIDAttr` with a custom assembly format that prints/parses to/from hex/binary. That could be a TODO if we don't want to spend time on that detail now.

test/Target/SubstraitPB/Import/literal.textpb
```diff
@@ -10,6 +10,60 @@
...
+            }
+          }
+          named_table {
+            names: "t1"
+          }
+        }
+      }
+      expressions {
+        literal {
+          uuid: "c48ffa9e-64f4-44cb-ae47-152b4e60e77b"
```
ingomueller-net, line 54: I don't think that this is how the Substrait spec intends to format the UUID value...

ingomueller-net (commented):
I think we need to revisit in what format we represent the literal value (see inline discussions).

test/Target/SubstraitPB/Import/literal.textpb
ingomueller-net: I am pretty sure that the `substrait-validator` interprets the literals in protobuf plans directly as 16-byte values, not as hex formatted values. See [parsing](https://github.com/substrait-io/substrait-validator/blob/68ba684d/rs/src/parse/expressions/literals.rs#L757-L781) logic.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
ingomueller-net: We *could* also represent the value internally as an `i128`; however, that'd require us to convert to and from 16-byte arrays, which I am not sure how easy it is. We can also leave that as a potential future optimization.

dshaaban01:
@ingomueller-net I restructured things and created a UUIDAttr instead of doing a TypedStrAttr. Let me know what you think!

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -949,6 +961,12 @@ SubstraitExporter::exportOperation(LiteralOp op) {
     intervalDaytoSecond->set_seconds(intervalSecond);
     literal->set_allocated_interval_day_to_second(
         intervalDaytoSecond.release());
+  } // `UUIDType`.
+  else if (auto uuidType = dyn_cast<UUIDType>(literalType)) {
+    llvm::APInt uuid = mlir::cast<UUIDAttr>(value).getValue().getValue();
+    std::string res(16, 0);
+    llvm::StoreIntToMemory(uuid, reinterpret_cast<uint8_t *>(res.data()), 16);
```
ingomueller-net, line 968: Nice! That's convenient :)

ingomueller-net (approved):
Yeah, I like that design. The LLVM functions are handy enough to not consider that a downside of using an `IntegerAttr`.

PR #50, opened 2025-01-06
chore: fix all test file spacing issues and remove redundant split input file lines

dshaaban01:
Saw some spacing errors in one test file and then decided I might as well go through all the test files and fixed all the spacing issues as well as removed redundant "split input file" lines from some of the tests.

### DIFF
--- a/test/Dialect/Substrait/fetch.mlir
+++ b/test/Dialect/Substrait/fetch.mlir
@@ -16,6 +16,8 @@ substrait.plan version 0 : 42 : 1 {
   }
 }
 
+// -----
+
 // CHECK-LABEL: substrait.plan
 // CHECK:         relation
 // CHECK:           %[[V0:.*]] = named_table
@@ -30,6 +32,8 @@ substrait.plan version 0 : 42 : 1 {
     yield %1 : tuple<si32>
   }
 }
+
+// -----
 
 // CHECK-LABEL: substrait.plan
 // CHECK:         relation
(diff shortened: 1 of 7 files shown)


test/Dialect/Substrait/literal-invalid.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: substrait-opt -verify-diagnostics -split-input-file %s
+// RUN: substrait-opt -verify-diagnostics %s
```
ingomueller-net, line 2:
```suggestion

```

test/Dialect/Substrait/call.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: substrait-opt -split-input-file %s \
```
ingomueller-net, line 1: Nit: I have ambiguous feelings about this change: it *does* remove a (currently) unused parameter but chance are that some (or even many?) of the files will eventually get another test cases added, in which case we'd want to add the flag again. Minor detail and not 100% clear, so feel free to decide either way.

ingomueller-net (approved):
LGTM, thanks a lot for the clean-up!

test/Dialect/Substrait/call.mlir
dshaaban01: I understand. I'll add them back (had to rebase now and it was annoying to have to add "split input file" stuff on test cases that grew.

dshaaban01:
Done.

PR #55, opened 2025-01-13
feat: show usage of builtin StringAttr with custon type [WIP]

ingomueller-net:


ingomueller-net:
Implemented in the two related PRs.

PR #56, opened 2025-01-13
feat: implement script that produces stats on spec coverage

ingomueller-net:
This PR adds a script that computes statistics on how much of the Substrait specification is currently covered by this project based on which of the protobuf fields of the spec occur in the "import" test cases. The PR also adds a new CI target that checks whether the markdown file holding the spec is up-to-date and fails with an error message indicating how to update that file in case that's necessary.

### DIFF
--- a/.github/workflows/coverage_stats.yml
+++ b/.github/workflows/coverage_stats.yml
@@ -1,0 +1,42 @@
+name: Coverage stats
+
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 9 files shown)


mortbopet (approved):
I don't have any comments, other than this is a very neat thing, and will be extremely useful in pinpointing exactly what needs work where, as well as making it clear for users what they can expect from the project.

PR #59, opened 2025-01-17
feat: elide attribute type with new `StaticallyTypedAttrInterface`

ingomueller-net:
This allows to omit the type of an attribute from the assembly if it can be inferred from the attribute value. For example, the type is redundant in `#substrait.timestamp<100us> : !substrait.timestamp`. The built-in `TypedAttrInterface`, however, forces the appearance of the type in the assembly. The new interface is almost identical but does not enforce it. The PR also makes the two timestamp, the date, and the time attributes implement that interface.

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -15,9 +15,24 @@ include "mlir/IR/AttrTypeBase.td"
 include "mlir/IR/BuiltinAttributeInterfaces.td"
 
 // Base class for Substrait dialect attribute types.
-class Substrait_Attr<string name, string typeMnemonic, list<Trait> traits = []>
+class Substrait_Attr<string name, string attrMnemonic, list<Trait> traits = []>
     : AttrDef<Substrait_Dialect, name, traits> {
-  let mnemonic = typeMnemonic;
+  let mnemonic = attrMnemonic;
+}
+
+// Base class for Substrait dialect attribute types that have a statically known
+// value type.
+class Substrait_StaticallyTypedAttr<
+      string name, string attrMnemonic, string typeName, list<Trait> traits = []>
+    : Substrait_Attr<name, attrMnemonic, traits # [
+      DeclareAttrInterfaceMethods<TypeInferableAttrInterface>
+    ]> {
+  let extraClassDeclaration = [{
+    /// Implement TypeInferableAttrInterface.
+    ::mlir::Type getType() {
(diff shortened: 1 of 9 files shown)


ingomueller-net:
@dshaaban01: Have a look at this PR. I managed to elide the attribute type from the assembly. I hope we'll be able to merge it soon, so consider starting any new work based on this PR if it's related.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -34,64 +49,44 @@ def Substrait_AdvancedExtensionAttr
...
-    ::mlir::Type getType() const {
-      return DateType::get(getContext());
-    }
-  }];
 }
 
-def Substrait_TimeAttr : Substrait_Attr<"Time", "time",
-    [TypedAttrInterface]> {
+def Substrait_TimeAttr :
+  Substrait_StaticallyTypedAttr<"Time", "time", "TimeType"> {
```
jpienaar, line 63: Micro nit: add additional spaces here to differentiate from body below (I normally just add 2 extra, else this and body looks at same scope ... I should really try out the new formatter)

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -34,64 +49,44 @@ def Substrait_AdvancedExtensionAttr
...
-  let extraClassDeclaration = [{
-    ::mlir::Type getType() const {
-      return TimeType::get(getContext());
-    }
-  }];
 }
 
-def Substrait_TimestampAttr : Substrait_Attr<"Timestamp", "timestamp",
-    [TypedAttrInterface]> {
+def Substrait_TimestampAttr
```
jpienaar, line 72: Nit: colon on this line to match the others.

jpienaar, line 72: Nit: colon on this line to match the others.

jpienaar (approved):
Nice, we could even have a variant of this upstream :)

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
ingomueller-net: Fixed. I tried `clang-format`, which seems to work on `.td` files as well, and it works. I have formatted the lines of this commit using it. However, I am unsure as to whether and how to apply or enforce it on the existing files. The diff is quite big and a big commit just for reformatting has disadvantages...

PR #61, opened 2025-01-17
build: add flag to treat warnings as errors and enable in CI

ingomueller-net:
This PR depends on and, therefor, includes #66.

This PR adds an option to our CMake files that allow treating compiler warnings as errors in the targets from this repository and enables that option in CI. This will enforce that we fix compiler warnings before merging PRs, which should help to keep our code clean.


jpienaar (approved):

ingomueller-net:
I think this will pass once #64 will be merged.

PR #70, opened 2025-01-24
feat: add `aggregate` op and extend functions accordingly

ingomueller-net:
~~This PR depends on and, therefor, includes #68.~~

This PR adds support for the `AggregateRel` from the Substrait spec in the form of the `aggregate` op. This is arguably the most complex op implemented so far. It has an optional enum argument that requires custom parsing, several optional regions that require custom parsing, an attribute that depends on the presence and contents of the regions and requires custom parsing to omit it in the common case, and return types that depend on the two regions and the attribute. What's more, the current version of the spec is such that it is almost impossibly to interpret "grouping sets" because it relies on protobuf message equality, which is something can protobuf does not offer. The current implementation, thus, implements a best effort by using op equality instead (but needs to run CSE during export to ensure op uniqueness). Finally, the PR also extends the `call` op to represent also `AggregateFunction` messages (in addition to `ScalarFunction` messages), which are used by the new `aggregate` op.

### DIFF
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -9,6 +9,7 @@
 #include "substrait-mlir/Dialect/Substrait/IR/Substrait.h"
 
 #include "mlir/IR/DialectImplementation.h" // IWYU pragma: keep
+#include "mlir/IR/PatternMatch.h"
 #include "mlir/Support/LogicalResult.h"
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/TypeSwitch.h" // IWYU pragma: keep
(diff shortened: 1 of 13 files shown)


lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -116,12 +117,311 @@ void printCountAsAll(OpAsmPrinter &printer, Operation *op, IntegerAttr count) {
...
+  return parser.emitError(loc)
+         << "has invalid aggregate invocation type specification: " << keyword;
+}
+
+void printAggregationInvocation(
+    OpAsmPrinter &printer, CallOp op,
+    AggregationInvocationAttr aggregationInvocation) {
+  if (aggregationInvocation &&
+      aggregationInvocation.getValue() != AggregationInvocation::all) {
+    // The whitespace printed here compensates the trimming of whitespace in
```
jpienaar, line 181: Can't one add it there ? E.g., use " " or some such? (just OOC, not blocker)

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -32,6 +33,30 @@ namespace pb = google::protobuf;
 
 namespace {
 
+// Copied from
+// https://github.com/llvm/llvm-project/blob/dea33c/mlir/lib/Transforms/CSE.cpp.
```
jpienaar, line 37: Could we just reuse it?

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -153,6 +184,154 @@ static mlir::FailureOr<mlir::Type> importType(MLIRContext *context,
   }
 }
 
+mlir::FailureOr<CallOp>
+importAggregateFunction(ImplicitLocOpBuilder builder,
+                        const AggregateFunction &message) {
+  MLIRContext *context = builder.getContext();
+  Location loc = UnknownLoc::get(context);
```
jpienaar, line 191: Why not use the location from the ImplicitLocOpBuilder?

jpienaar: (same below)

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -153,6 +184,154 @@ static mlir::FailureOr<mlir::Type> importType(MLIRContext *context,
...
+        if (failed(exprOp))
+          return failure();
+
+        // Create or look-up reference.
+        auto [it, hasInserted] = groupingExprOps.try_emplace(exprOp.value());
+
+        // If it's a new expression, assign new reference.
+        if (hasInserted) {
+          it->second = groupingExprOps.size() - 1;
+          groupingExprValues.emplace_back(exprOp.value()->getResult(0));
```
jpienaar, line 290: And these will always have a single result?

jpienaar (approved):

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: I haven't found a way that makes me perfectly happy: if I add it there, I get *two* whitespaces in case this custom directive doesn't print anything. So now the declarative format suppresses a whitespace before this directive, then (1) if the directive prints something, it first prints the whitespace, then that thing, (2) otherwise it doesn't print anything, and in both cases, the declarative format adds another whitespace.

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: Unfortunately, no, at least not currently. It's in the `cpp` file, so we'd have to move it upstream to some header to make it accessible.

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: Good idea! Fixed in latest commit.

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: Yes, but currently only by convention. This is bothering for some time but I haven't gotten around to trying to improve things again. Ideally, I would like to attach the `OpTrait::OneResult` to the `RelOpInterface` and the `ExpressionOpInterface` but I didn't immediate manage to do that (and I am not sure if we can attach a trait to an *interface*). The best alternative I came up with is to add a `getResult()` method to the interface that simply does `return getResult(0)`.

PR #76, opened 2025-01-31
feat: implement `cast` op for `Cast` expression

ingomueller-net:
This PR depends on and, therefor, includes #75.

This PR implements the `cast` op, which corresponds to the `Cast` expression in Substrait. This is one of the few remaining message types we need to support to get a first TPC-H query round-tripped.


ingomueller-net:
For lack of other place to keep track of this, the following message types/fields are still missing for the [plan](https://github.com/substrait-io/consumer-testing/blob/cd2372f/substrait_consumer/snapshots/producer/integration/tpch/q06-isthmus_plan.json) of TPC-H Q6 of Isthmus:

* `Type.Struct`
* `Type.Decimal`
* `nullability` in various places
* `Expression.Literal.fixed_char`
* `AggregateFunction.phase`
* `Expression.ReferenceSegment.StructField.field`

@dshaaban01: you may also find that interesting.

jpienaar (approved):

PR #77, opened 2025-02-03
feat: implement `aggregation_phase` of `call` op

ingomueller-net:
~~This PR is based on and, therefor, includes #70.~~

This PR adds the `aggregation_phase` to the `call` op, which corresponds to the `phase` field of the `AggregationFunction` message. The PR also revisits the assembly format of the `aggregation_invocation` attribute of the `call` op and now handles the two jointly such that either of the two can be ommitted if it has the default value. Finally, the PR revisits the LIT tests such that both aggregation detauls are tested jointly and explicitly instead of being piggy-backed onto the remaining tests. This is the last non-type feature we need to run TPC-H Q6 as produced by Isthmus.

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
@@ -371,6 +371,12 @@ def Substrait_LiteralOp : Substrait_ExpressionOp<"literal", [
 
 def Substrait_CallOp : Substrait_ExpressionOp<"call", [
     DeclareOpInterfaceMethods<SymbolUserOpInterface>,
+    PredOpTrait<"'aggregation_phase' and 'aggregation_invocation' must either "
+                "both be set or both not be set",
+                CPred<[{ ::llvm::all_equal(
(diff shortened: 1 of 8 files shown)


include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -371,16 +371,27 @@ def Substrait_LiteralOp : Substrait_ExpressionOp<"literal", [
...
-    (otherwise optional) aggregate or window-related attributes. For aggregate
-    functions, the invocation type is omitted from the custom assembly if it is
-    set to `all`.
+    (otherwise optional) aggregate or window-related attributes.
+
+    For aggregate functions, the aggregation phase is ommitted if it is set to
+    `initial_to_result`; the invocation type is omitted from the custom assembly
+    if it is set to `all`. If only `unspecified` is provided, the aggregation
+    phase will be set to that value and the invocation type to `all`. The
+    printer avoids that case by always printing the attribute value of the other
```
mortbopet, line 393: of->if?

mortbopet (approved):

PR #78, opened 2025-02-04
feat: implement real `relation` type

ingomueller-net:
~~Once rebased onto #87, this PR will need llvm/llvm-project#127518 to make `pyright` happy.~~

This PR introduces a new `RelationType` as the operands and result types of `RelOp`s, where previously `tuple` had been used as a placeholder. While this changes little in the structure of the IR, it (1) underlines the subtle difference between the relation, a *container* of records, and the records themselves, which had *both* been represented as `tuple`s before and (2) it allows defining an even shorter type name, `rel`, for that type. The latter is implemented using a custom printer and parser for these "pseudo type names" (which are really keywords for the printer and parser), which lays the basis for similar short-hand forms for some of our other types with long type names.

### DIFF
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -398,6 +398,46 @@ void printFixedBinaryLiteral(AsmPrinter &printer, StringAttr value,
   printer << value;
 }
 
+static StringRef getTypeKeyword(Type type) {
+  return TypeSwitch<Type, StringRef>(type)
+      .Case<RelationType>([&](Type) { return "rel"; })
+      .Default([](Type) -> StringRef {
+        llvm_unreachable("unexpected 'llvm' type kind");
+      });
+}
+
+ParseResult parseSubstraitType(AsmParser &parser, Type &type) {
+  SMLoc loc = parser.getCurrentLocation();
+
+  // Try parsing any MLIR type in full form.
+  OptionalParseResult result = parser.parseOptionalType(type);
+  if (result.has_value()) {
+    if (failed(result.value()))
+      return failure();
+    return success();
+  }
+
+  // If no type is found, expect a short Substrait type keyword.
+  StringRef keyword;
+  if (failed(parser.parseKeyword(&keyword)))
+    return failure();
+
+  // Dispatch parsing to type class depending on keyword.
+  type = StringSwitch<function_ref<Type()>>(keyword)
+             .Case("rel", [&] { return RelationType::parse(parser); })
+             .Default([&] {
+               parser.emitError(loc) << "unknown Substrait type: " << keyword;
+               return Type();
+             })();
+  return success(type != Type());
+}
+
+void printSubstraitType(AsmPrinter &printer, Operation *op, Type type) {
+  printer << getTypeKeyword(type);
+  llvm::TypeSwitch<Type>(type).Case<RelationType>(
+      [&](auto type) { type.print(printer); });
+}
+
 Type RelationType::parse(AsmParser &parser) {
   // Parse `<` literal.
   if (failed(parser.parseLess()))
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -401,16 +401,14 @@ void printFixedBinaryLiteral(AsmPrinter &printer, StringAttr value,
 static StringRef getTypeKeyword(Type type) {
   return TypeSwitch<Type, StringRef>(type)
       .Case<RelationType>([&](Type) { return "rel"; })
-      .Default([](Type) -> StringRef {
-        llvm_unreachable("unexpected 'llvm' type kind");
-      });
+      .Default([](Type t) -> StringRef { return ""; });
 }
 
-ParseResult parseSubstraitType(AsmParser &parser, Type &type) {
+ParseResult parseSubstraitType(AsmParser &parser, Type &valueType) {
   SMLoc loc = parser.getCurrentLocation();
 
   // Try parsing any MLIR type in full form.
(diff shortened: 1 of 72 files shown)


ingomueller-net:
I am still unsure about what assembly format is best. We have the following possibilities:

```
%0 = named_table @t1 as ["a"] : <si32>  // no typename by default
%0 = named_table @t1 as ["a"] : !substrait.relation<si32>  // always fully qualified name
%0 = named_table @t1 as ["a"] : relation<si32>  // short pseudo-typename
```

If no type name is the default, we have to rely on the reader's knowledge that the ops always return relations. If we always print the fully qualified name, some ops get quite verbose (set ops can have several operands plus the result, which would all repeat that it's a relation). The short pseudo-typename would really be a keyword, which (1) would need to be defined for each op and (2) wouldn't color as a type in IDEs. Any opinions, @jpienaar or @dshaaban01?

ingomueller-net:
@mortbopet: Any opinion from your side?

mortbopet:
Well, first, i support adding a relation type to get out of the whole aliasing-with-tuples-in-IR issue. But it is a shame that we can't re-use the tuple type as an internal storage mechanism, and have to copy the type storage struct - which appears to me as non-trivial.
Is it not possible to have an implementation that keeps the custom printer/parser, but has a `TupleType` as parameter? That should take care of inferring a type storage, but we'd then still provide the custom builders for API neatness.

Re. assembly format, given the reasons that you highlight, i'd be in favor of the first (no typename). This would also be in line with substrait operations generally having elided their dialect name (`getDialectNamespace`). But mainly due to the size of the IR; i've already experienced IR bloat through `substrait.decimal` types (also a long name!) so adding in even more text would clearly worsen this.

ingomueller-net:
I though a bit more about the "only a pseudo-type" issue. I think it can be acceptable, and I think it'll provide a better trade-off in other situations as well, namely to reduce the redundancy of nested types, including the field types of `!substrait.relation`. We have discussed this [here](https://github.com/substrait-io/substrait-mlir-contrib/pull/121#discussion_r2031389919) for the list type. Concretely, I am thinking to use pseudo-types similar to how the LLVM dialect handles this upstream [here](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/LLVMIR/IR/LLVMTypeSyntax.cpp):

```cpp
static StringRef getTypeKeyword(Type type) {
  return TypeSwitch<Type, StringRef>(type)
      .Case<LLVMVoidType>([&](Type) { return "void"; })
      .Case<LLVMPPCFP128Type>([&](Type) { return "ppc_fp128"; })
      .Case<LLVMTokenType>([&](Type) { return "token"; })
     // ...
      .Default([](Type) -> StringRef {
        llvm_unreachable("unexpected 'llvm' type kind");
      });
}

void mlir::LLVM::detail::printType(Type type, AsmPrinter &printer) {
  // ....
  printer << getTypeKeyword(type);
  llvm::TypeSwitch<Type>(type)
      .Case<LLVMPointerType, LLVMArrayType, LLVMFunctionType, LLVMTargetExtType,
            LLVMStructType>([&](auto type) { type.print(printer); });
}

static Type dispatchParse(AsmParser &parser, bool allowAny = true) {
  // ...
  return StringSwitch<function_ref<Type()>>(key)
      .Case("void", [&] { return LLVMVoidType::get(ctx); })
      .Case("ppc_fp128", [&] { return LLVMPPCFP128Type::get(ctx); })
      .Case("token", [&] { return LLVMTokenType::get(ctx); })
      // ...
      .Default([&] {
        parser.emitError(keyLoc) << "unknown LLVM type: " << key;
        return Type();
      })();
}
```

That could lead to an assembly format like this:

```mlir
  relation {
    %0 = named_table @t1 as ["a"] : rel<timestamp_tz>
    %1 = named_table @t2 as ["b"] : rel<timestamp_tz>
    %2 = join inner %0, %1 : rel<timestamp_tz>, rel<timestamp_tz> -> rel<timestamp_tz, timestamp_tz>
    yield %2 : rel<timestamp_tz>
  }
}
```

Would that be an acceptable trade-off?

ingomueller-net:
> I though a bit more about the "only a pseudo-type" issue.

I implemented part of this idea in the last commit of this PR and #131 stacked on top of it in order to see betterhow this could look like. @dshaaban01, @mortbopet: what are your impressions?

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -397,6 +398,79 @@ void printFixedBinaryLiteral(AsmPrinter &printer, StringAttr value,
   printer << value;
 }
 
+static StringRef getTypeKeyword(Type type) {
+  return TypeSwitch<Type, StringRef>(type)
+      .Case<RelationType>([&](Type) { return "rel"; })
+      .Default([](Type) -> StringRef {
+        llvm_unreachable("unexpected 'llvm' type kind");
```
mortbopet, line 405: is `llvm` intended here?

mortbopet (commented):
@ingomueller-net sorry for the delay. Glad to see that there was already a version upstream for handling this kind of scenario. I'm on-board with the shorthand version (#131).

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: Good catch! No, not intended. Will fix in the next iteration.

ingomueller-net:
> @ingomueller-net sorry for the delay. Glad to see that there was already a version upstream for handling this kind of scenario. I'm on-board with the shorthand version (#131).

OK, cool, thanks for the feedback and no worries about delays -- I am have quite a few distractions these days myself... OK, will try to finish this iteration at some point next week.

mortbopet (approved):

PR #74, opened 2025-01-30
feat: update LLVM to latest version as of Jan. 31, 2025 (956cfa69b15) 

ingomueller-net:
This PR updates the LLVM git submodule to the latest version as of
January 31, 2025, namely llvm-project/llvm@956cfa69b15 and adapts
the code base to the new version.

PR #80, opened 2025-02-06
chore: remove type of date/time attribute from test

dshaaban01:
Removed types of date/time attributes from the literal test and Export/literal tests to align with the timestamp tests. I noticed that these attributes were still present in date/time but not in timestamp tests - was this an oversight or intentional? Let me know if they were intentionally kept and will close the PR.

### DIFF
--- a/test/Dialect/Substrait/literal.mlir
+++ b/test/Dialect/Substrait/literal.mlir
@@ -16,7 +16,7 @@ substrait.plan version 0 : 42 : 1 {
     %0 = named_table @t1 as ["a"] : tuple<si1>
     %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.time> {
     ^bb0(%arg : tuple<si1>):
-      %time = literal #substrait.time<200000000us> : !substrait.time
+      %time = literal #substrait.time<200000000us> 
       yield %time : !substrait.time
     }
     yield %1 : tuple<si1, !substrait.time>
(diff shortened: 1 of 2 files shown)


dshaaban01:
@ingomueller-net 

Same weird download error as in #48 ...

test/Dialect/Substrait/literal.mlir
```diff
@@ -16,7 +16,7 @@ substrait.plan version 0 : 42 : 1 {
     %0 = named_table @t1 as ["a"] : tuple<si1>
     %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.time> {
     ^bb0(%arg : tuple<si1>):
-      %time = literal #substrait.time<200000000us> : !substrait.time
+      %time = literal #substrait.time<200000000us>
```
ingomueller-net, line 19: Ultra nit: remove the trailing whitespaces please.

```suggestion
      %time = literal #substrait.time<200000000us>
```

ingomueller-net (approved):
No, not intentional, good catch! Thanks!

ingomueller-net (approved):

ingomueller-net:
Remember that #74 updated LLVM. I think it'd be best to rebase your PRs to main such that (1) we don't thrash the `ccache` during CI and (2) we don't run into surprise breakages due to the rebase (although that's unlikely).

ingomueller-net:
I rebased locally in my machine and ran the tests, so need to rebase this one.

PR #81, opened 2025-02-06
feat: introduce `fixedchar<L>` type

dshaaban01:
This PR introduces the `fixedchar<L>` type, which represents a fixed-length Unicode string of L characters.


dshaaban01:
Also failing for same reasons as in other PRs.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -362,8 +362,8 @@ def Substrait_LiteralOp : Substrait_ExpressionOp<"literal", [
     ```
   }];
   // TODO(ingomueller): extend to other types.
-  let arguments = (ins Substrait_AtomicAttribute:$value);
-  let results = (outs Substrait_AtomicType:$result);
+  let arguments = (ins AnyAttrOf<[Substrait_AtomicAttribute,Substrait_ContainerAttribute]>:$value);
```
ingomueller-net, line 365: Maybe it would make sense to add a `def` to `SubstraitAttrs.td` called something like `Substrait_LiteralAttribute`?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -362,8 +362,8 @@ def Substrait_LiteralOp : Substrait_ExpressionOp<"literal", [
...
+  let results = (outs AnyTypeOf<[Substrait_AtomicType,Substrait_ContainerType]>:$result);
```
ingomueller-net, line 366: Same for the types. Maybe `Substrait_ExpressionType`? Not 100% happy with that... is it really the "type of the expression" or rather the "type of the *result* of the expression"?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -189,9 +189,20 @@ def Substrait_AtomicAttributes {
...
+  list<Attr> attrs = [
+    TypedStrAttr<Substrait_FixedCharType>, // FixedChar
+  ];
+}
+
 /// Attribute of one of the currently supported atomic types.
 def Substrait_AtomicAttribute : AnyAttrOf<Substrait_AtomicAttributes.attrs>;
 
+/// Attribute of one of the currently supported compound types.
+def Substrait_ContainerAttribute : AnyAttrOf<Substrait_ContainerAttributes.attrs>;
```
ingomueller-net, line 204: You may not have revisited that yet but remember our discussion on "container" vs. "parametrized"...

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -114,7 +132,7 @@ def Substrait_AnyType : Substrait_Type<"Any", "any"> {
 def Substrait_AtomicType : AnyTypeOf<Substrait_AtomicTypes.types>;
 
 /// Any container type, i.e., structs, maps, lists, and nestings thereof.
-def Substrait_ContainerType : NestedTupleOf<Substrait_AtomicTypes.types>;
+def Substrait_ContainerType : AnyTypeOf<Substrait_ContainerTypes.types#[NestedTupleOf<Substrait_AtomicTypes.types>]>;
```
ingomueller-net, line 135: Can you reformat that to fit into the 80-character limit?

dshaaban01:
Add verifier.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
dshaaban01: To make sure I remember correctly: I will simply add a new "parametrized type" and include any and all types (from the atomic AND compound type list - as in the spec) that require a parameter? Right?

dshaaban01: @ingomueller-net Could you elaborate a bit more to jog my memory? I wrote a previous question/comment that I think was incorrect after doing some more research online, so I deleted .

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
dshaaban01: What about `Substrait_ResultExpressionType`?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -203,9 +203,24 @@ def Substrait_AtomicAttributes {
...
+
+/// Attribute of one of the currently supported atomic or compound types.
+def Substrait_LiteralAttribute : AnyAttrOf<[Substrait_AtomicAttribute,Substrait_ContainerAttribute]>;
```
ingomueller-net, line 221: Nit:
```suggestion
def Substrait_LiteralAttribute : AnyAttrOf<[Substrait_AtomicAttribute, Substrait_ContainerAttribute]>;
```

Also check the line length; I can't see it here on Github.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -203,9 +203,24 @@ def Substrait_AtomicAttributes {
...
+
+
```
ingomueller-net, line 223: Nit:
```suggestion
```

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -33,17 +33,27 @@ def Substrait_DateType : Substrait_Type<"Date", "date"> {
   }];
 }
 
+def Substrait_FixedCharType : Substrait_Type<"FixedChar", "fixed_char"> {
+  let summary = "Substrait fixed-length char type";
+  let description = [{
+    This type represents a substrait fixed-length unicode string of L characters.
```
ingomueller-net, line 39:
```suggestion
    This type represents a substrait fixed-length unicode string of `L` characters.
```

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -108,6 +118,15 @@ def Substrait_AtomicTypes {
   ];
 }
 
+/// Currently supported compound types, listed in order of substrait specification.
+/// These correspond directly to the types in
+/// https://github.com/substrait-io/substrait/blob/main/proto/substrait/type.proto.
+def Substrait_ContainerTypes {
```
ingomueller-net, line 124: Remember to adapt this name if you adapt the other one.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -124,17 +143,29 @@ def Substrait_AnyType : Substrait_Type<"Any", "any"> {
...
 ]>;
 
+/// One of the types of the result of an expression.
+def Substrait_ResultExpressionType :
+    AnyTypeOf<
+        [
+            Substrait_AtomicType,
+            Substrait_ContainerType
+        ]
+    >;
```
ingomueller-net, line 164: Nit:
```suggestion
def Substrait_ResultExpressionType : AnyTypeOf<[
    Substrait_AtomicType,
    Substrait_ContainerType,
  ]>;
```

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
ingomueller-net, line 221: I'd say that what we call "literal" here refers to (exactly? roughly?) the same concept as the one we call "expression result" below. We should spend some time finding a good name. Here are some more thoughts.

* "Literal" fits pretty well for the "attribute" because all possible "types" should be possible attribute values for the `LiteralOp`. However, I'd say it doesn't fit as well for types outside the context of the `LiteralOp`. The result of `FieldReferenceOp` isn't a literal after all.
* "Value": could be confused with MLIR's SSA values, which also refer to relations.
* "Field" (or "Column") would refer to what relations consist of using terms that are also used in the Substrait spec. I'd argue that it fits well in `FieldType`; I am not 100% convinced of `FieldAttribute`. But maybe it makes sense that the `LiteralAttribute` can hold `FieldType`s?
* I think we considered `ExpressionResultType`; you use `ResultExpressionType`. How well do those fit? In general, quite well: it's the type of the result of `Expression` ops. However, in a `LiteralOp`, technically, we don't store an expression result... Even the fields in a `ReadRel` aren't result of expressions.

After writing this up, I think my favorites are `FieldType` and `LiteralAttribute`. WDYT?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -124,17 +143,29 @@ def Substrait_AnyType : Substrait_Type<"Any", "any"> {
...
+
 /// Placeholder for a proper relation type, the result of any `RelOpInterface`
 /// op.
 // TODO(ingomueller): Transform this into a proper relation type.
-def Substrait_Relation : NestedTupleOf<Substrait_AtomicTypes.types>;
+def Substrait_Relation : NestedTupleOf<Substrait_AtomicTypes.types#Substrait_ContainerTypes.types>;
```
ingomueller-net, line 169: Does this still fit into 80 characters?

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -68,6 +68,13 @@ LogicalResult AdvancedExtensionAttr::verify(
   return success();
 }
 
+LogicalResult mlir::substrait::FixedCharType::verify(
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError, int32_t length) {
+  if (length < 1 || length > 2147483647)
```
ingomueller-net, line 73: Where does this constant come from? Is it `2^31 - 1`? If so, use that. Also: `int32_t` doesn't hold larger values that that, so the test is always true.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -68,6 +68,13 @@ LogicalResult AdvancedExtensionAttr::verify(
   return success();
 }
 
+LogicalResult mlir::substrait::FixedCharType::verify(
```
ingomueller-net, line 71: Note that this verifies the *type* and ensures that its parameter `L` has a valid value. We also want to test the *attribute*, though, and ensure that the string that it holds has exactly length `L`. I think we need to `def` an attribute for that purpose and implement a C++ verifier with that check.

ingomueller-net (commented):
We are getting there :) There are still some things that I think need work. I hope that this isn't too frustrating! This PR lays out a couple of fundamental concepts, so it isn't surprising that we are doing a few iterations.

lib/Dialect/Substrait/IR/Substrait.cpp
dshaaban01: I was hesitant to add an attribute definition because FixedCharType<6> with the value "hello" is not the same as FixedCharType<5> with the same value. Since the type encodes the length, how can I extract the length (L) directly from the type rather than relying on an attribute?

dshaaban01: Will play around and see if I can somehow hook the `L` from the type to the attribute type.

dshaaban01:
@ingomueller-net After some mental gymnastics, I think what I pushed right now is a lot closer to our ideal! Some main points: 

- I now have a `CompoundAttribute` instead of `ContainerAttribute`
- Changed name of `Substrait_ResultExpressionType` to `Substrait_ExpressionResultType` as I thought this made more sense. 
- I added `Substrait_CompoundType`: I did NOT replace `Substrait_ContainerType` as we discussed that Compound and Container are two different things. I thought I would add `Substrait_CompoundType` to the definition of `Substrait_ContainerType` in a separate PR, since a container type can hold a compound type (a struct with a field that is a compound type such as FixedChar<L>)? WDYT? We can discuss further tomorrow :)) 
- Added a `Substrait_FixedCharAttr` that takes as parameters a value and a type (see above comment on why I thought this was necessary). Note that this Attribute is not a `Substrait_StaticallyTypedAttr` but is a `TypedAttrInterface` - this is because it's no longer statically type and needs the parameter to know the type - there is no single type.
- By adding `AttributeSelfTypeParameter` around the type parameter, I avoid having to write `#substrait.fixed_char<"hello", <5>>` in my tests and can instead have `#substrait.fixed_char<"hello">`.
- Due to the addition of `AttributeSelfTypeParameter`, I had to alter the verifier to accommodate for the fact that the `type` parameter is not a `FixedCharType` as it originally was, but rather now an mlir `Type`.

I think this about sums it up!! I will now take the same approach for my reimplementation of `VarChar<L>`.

lib/Dialect/Substrait/IR/Substrait.cpp
dshaaban01: I remembered that I had made an attempt to do this for VarChar and wrote a comment about it [here](https://github.com/substrait-io/substrait-mlir-contrib/pull/82#issuecomment-2678780595).

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -70,6 +70,18 @@ def Substrait_DateAttr
   let assemblyFormat = [{ `<` $value `>` }];
 }
 
+def Substrait_FixedCharAttr
+    : Substrait_Attr<"FixedChar", "fixed_char", [TypedAttrInterface]> {
+  let summary = "Substrait fixed-length char attributetype";
+  let description = [{
+    This type represents a substrait fixed-length unicode string of `L`
+    characters attribute type.
```
ingomueller-net, line 78: This is by accident, right?
```suggestion
    characters.
```

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -203,9 +215,25 @@ def Substrait_AtomicAttributes {
   ];
 }
 
+/// Attributes of currently supported compound types, listed in order of
+/// substrait specification.
+def Substrait_CompoundAttributes {
```
ingomueller-net, line 220: I argued for using the term "compound" but I think I got mislead by my limited English. "compound" is sort of synonym to "container", no? A fixed char doesn't "compound" anything, does it? Or at least a decimal doesn't "compound" anything. I think what we said at one point is that these are *parametrized*, no?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -123,18 +142,30 @@ def Substrait_AnyType : Substrait_Type<"Any", "any"> {
...
+/// One of the currently supported atomic, compound or nested types.
 def Substrait_FieldType : AnyTypeOf<[
   Substrait_AtomicType,
-  Substrait_ContainerType
+  Substrait_CompoundType,
+  Substrait_ContainerType,
 ]>;
 
+/// One of the types of the result of an expression.
+def Substrait_ExpressionResultType : AnyTypeOf<[
```
ingomueller-net, line 159: Let's review the discussion for this name in our meeting in a bit.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -68,6 +68,19 @@ LogicalResult AdvancedExtensionAttr::verify(
   return success();
 }
 
+LogicalResult mlir::substrait::FixedCharAttr::verify(
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError, StringAttr value,
+    Type type) {
+  FixedCharType fixedCharType = mlir::dyn_cast<FixedCharType>(type);
+  if (fixedCharType == nullptr)
+    return emitError() << "expected a fixed char type";
```
ingomueller-net, line 76: Nit: use the type name from the assembly format and give details on where it was expected.
```suggestion
    return emitError() << "expected 'type' to be a 'fixed_char'";
```

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -68,6 +68,19 @@ LogicalResult AdvancedExtensionAttr::verify(
...
+  int32_t value_length = value.size();
+  if (value_length != fixedCharType.getLength())
+    return emitError() << "value length must be" << fixedCharType.getLength()
+                       << "characters.";
```
ingomueller-net, line 80: Nit: spaces.
```suggestion
    return emitError() << "value length must be " << fixedCharType.getLength()
                       << " characters.";
```

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -675,6 +677,12 @@ importLiteral(ImplicitLocOpBuilder builder,
     auto attr = UUIDAttr::get(context, integer_attr);
     return builder.create<LiteralOp>(attr);
   }
+  case Expression::Literal::LiteralTypeCase::kFixedChar: {
+    auto attr = FixedCharAttr::get(
+        context, StringAttr::get(context, message.fixed_char()),
+        FixedCharType::get(context, message.fixed_char().size()));
```
ingomueller-net, line 683: Nit: factor out the `StringAttr` and the `FixedCharType` into dedicated lines.

test/Dialect/Substrait/literal.mlir
```diff
@@ -1,12 +1,36 @@
...
+// CHECK-NEXT:      yield %[[V2]] : !substrait.fixed_char<5>
+// CHECK-NEXT:    }
+// CHECK-NEXT:    yield %[[V1]] : tuple<si1, !substrait.fixed_char<5>>
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a"] : tuple<si1>
+    %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.fixed_char<5>> {
+    ^bb0(%arg : tuple<si1>):
+      %fixed_char = literal #substrait.fixed_char<"hello"> : !substrait.fixed_char<5>
```
ingomueller-net, line 19: There's some redundancy here: The attribute name `#substrait.fixed_char` and the length of the string already fully define the type `!substrait.fixed_char<5>`. Do we want to remove it? I think that this could be done with our `TypeInferableAttrInterface`. If we don't do it now, please leave a TODO.

ingomueller-net (approved):
We are close :) Only minor comments now. Thanks for persisting!

dshaaban01:
Update: everything done except for correct implementation of type inference. Still on it.

ingomueller-net (approved):
LGTM! A few minor nits but *really, really* minor now :)

Also, I am going to merge #108 in a second. *If you can do it in a few minutes*, could you rebase/merge and add the `Decimal` type and attribute to the corresponding lists? Otherwise, we'll do a new PR that does only that.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -203,8 +221,20 @@ def Substrait_AtomicAttributes {
   ];
 }
 
+/// Attributes of currently supported compound types, listed in order of
+/// substrait specification.
```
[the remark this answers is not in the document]
dshaaban01: Oh forgot the comments, sorry about that!

dshaaban01:
@ingomueller-net 

Rebased and comments addressed.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -70,6 +70,27 @@ def Substrait_DateAttr
...
+    characters.
+  }];
+  let parameters = (ins "StringAttr":$value,
+      AttributeSelfTypeParameter<"FixedCharType">:$type);
+  let assemblyFormat = [{ `<` $value `>` }];
+  let genVerifyDecl = 1;
+  let extraClassDeclaration = [{
+    /// Implement TypeInferableAttrInterface.
+    ::mlir::Type getType() {
+      return ::mlir::substrait::FixedCharType::get(getContext(), getValue().size());
```
ingomueller-net, line 89: Didn't comment earlier but: nice! :)

dshaaban01:
Yeah!! First merge button I've clicked!! :conference

PR #83, opened 2025-02-06
feat: introduce `fixedbinary<L>` type

dshaaban01:
This PR is based on and, therefor, contains #82.

This PR introduces the compound type `fixedBinary<L>` representing a binary string of L bytes.

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -120,6 +120,16 @@ def Substrait_FixedCharAttr
   }];
 }
 
+def Substrait_FixedBinaryAttr : Substrait_Attr<"FixedBinary", "fixed_binary",
+  [TypedAttrInterface ]> {
+  let summary = "Substrait fixed-length binary type";
+  let description = [{
+    This type represents a substrait binary string of L bytes.
+  }];
+  let parameters = (ins "StringAttr":$value, "FixedBinaryType":$type);
+  let assemblyFormat = "`<` custom<AddRemovePadding>($value, $type) `>`";
+}
+
 def Substrait_IntervalDaySecondAttr
     : Substrait_StaticallyTypedAttr<"IntervalDaySecond", "interval_day_second",
                                     "IntervalDaySecondType"> {
@@ -271,6 +281,7 @@ def Substrait_ParametrizedAttributes {
   list<Attr> attrs = [
     Substrait_FixedCharAttr, // FixedChar
     Substrait_VarCharAttr, // VarChar
+    Substrait_FixedBinaryAttr, // FixedBinary
     Substrait_DecimalAttr, // Decimal
   ];
 }
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -128,7 +128,7 @@ def Substrait_FixedBinaryAttr : Substrait_Attr<"FixedBinary", "fixed_binary",
   }];
   let parameters = (ins "StringAttr":$value, "FixedBinaryType":$type);
   let genVerifyDecl = 1;
-  let assemblyFormat = "`<` custom<AddRemovePadding>($value, $type) `>`";
+  let assemblyFormat = "`<` custom<FixedBinaryTypeRemoveLength>($value, $type) `>`";
 }
 
 def Substrait_IntervalDaySecondAttr
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -128,7 +128,7 @@ def Substrait_FixedBinaryAttr : Substrait_Attr<"FixedBinary", "fixed_binary",
   }];
   let parameters = (ins "StringAttr":$value, "FixedBinaryType":$type);
   let genVerifyDecl = 1;
-  let assemblyFormat = "`<` custom<FixedBinaryTypeRemoveLength>($value, $type) `>`";
+  let assemblyFormat = "`<` custom<FixedBinaryLiteral>($value, $type) `>`";
 }
 
 def Substrait_IntervalDaySecondAttr
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -121,12 +121,13 @@ def Substrait_FixedCharAttr
 }
 
 def Substrait_FixedBinaryAttr : Substrait_Attr<"FixedBinary", "fixed_binary",
-  [TypedAttrInterface ]> {
+    [TypedAttrInterface]> {
   let summary = "Substrait fixed-length binary type";
   let description = [{
     This type represents a substrait binary string of L bytes.
   }];
   let parameters = (ins "StringAttr":$value, "FixedBinaryType":$type);
+  let genVerifyDecl = 1;
   let assemblyFormat = "`<` custom<AddRemovePadding>($value, $type) `>`";
 }
 
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -369,8 +369,9 @@ void printDecimalNumber(AsmPrinter &printer, DecimalType type,
   printer << "P = " << type.getPrecision() << ", S = " << type.getScale();
 }
 
-ParseResult parseFixedBinaryTypeRemoveLength(AsmParser &parser, StringAttr &value,
-                                  FixedBinaryType &type) {
+ParseResult parseFixedBinaryTypeRemoveLength(AsmParser &parser,
+                                             StringAttr &value,
+                                             FixedBinaryType &type) {
   std::string valueStr;
 
   // Parse fixed binary value as quoted string.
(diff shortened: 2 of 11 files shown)


include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -91,6 +91,16 @@ def Substrait_FixedCharAttr
   }];
 }
 
+def Substrait_FixedBinaryAttr : Substrait_Attr<"FixedBinary", "fixed_binary",
+  [TypedAttrInterface ]> {
```
ingomueller-net, line 95: Nits: indent one step more and remove stray space.
```suggestion
    [TypedAttrInterface]> {
```

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -355,6 +355,47 @@ void printDecimalNumber(AsmPrinter &printer, DecimalType type,
...
+                                  FixedBinaryType &type) {
+  // Parse fixed binary value as quoted string.
+  std::string valueStr;
+  if (parser.parseString(&valueStr))
+    return failure();
+
+  // Parse `L = <length>`.
+  uint32_t length;
+  if (parser.parseComma() || parser.parseInteger(length))
+    return failure();
```
ingomueller-net, line 368: These *could* be merged and done in one `if` statement...

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -355,6 +355,47 @@ void printDecimalNumber(AsmPrinter &printer, DecimalType type,
...
+
+  // Create `FixedBinaryType`.
+  auto emitError = [&]() {
+    return parser.emitError(parser.getCurrentLocation());
+  };
+  if (!(type = FixedBinaryType::getChecked(emitError, parser.getContext(),
```
ingomueller-net, line 374: Nit: Factor the context out into a dedicated variable.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -355,6 +355,47 @@ void printDecimalNumber(AsmPrinter &printer, DecimalType type,
...
+                                           length)))
+    return failure();
+
+  // Pad or trim the value to match the fixed length.
```
ingomueller-net, line 378: I am not sure if I like the padding. First, we can also create these attributes programmatically, for example during import, and those will not be padded. The padding of those wouldn't be removed during printing, but parsing that back in would then remove it, so they'd round-trip to a *different* attribute. In order to make that consistent, we'd have to verify that there is no padding and create convenience builders that remove the padding from arbitrary strings. Second, I only see a small benefit (a slightly more concise assembly for values that happen to end with `\0`) that also has the danger of bringing confusion (a mismatch between the apparent value length and the length specified in the type).

ingomueller-net (approved):
Nice! More progress :) LGTM!

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -80,6 +80,17 @@ LogicalResult mlir::substrait::FixedCharAttr::verify(
...
 }
 
+LogicalResult mlir::substrait::FixedBinaryAttr::verify(
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError, StringAttr value,
+    FixedBinaryType type) {
+  FixedBinaryType fixedBinaryType = mlir::dyn_cast<FixedBinaryType>(type);
+  int32_t value_length = value.size();
+  if (fixedBinaryType != nullptr && value_length != fixedBinaryType.getLength())
+    return emitError() << "value length must be " << fixedBinaryType.getLength()
+                       << " characters.";
```
ingomueller-net, line 90: Shouldn't we *fail* if `fixcedBinaryType` is a `nullptr`?
```suggestion
  if (fixedBinaryType == nullptr || value_length != fixedBinaryType.getLength())
    return emitError() << "value length must be " << fixedBinaryType.getLength()
                       << " characters.";
```
We should either (1) convince ourselves that `type` is always a `FixedBinaryType`, use a `cast` instead of a `dyn_cast` above and remove the test here or (2) use a dedicated test and error message for the `dyn_cast`.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -121,12 +121,13 @@ def Substrait_FixedCharAttr
...
 def Substrait_FixedBinaryAttr : Substrait_Attr<"FixedBinary", "fixed_binary",
-  [TypedAttrInterface ]> {
+    [TypedAttrInterface]> {
   let summary = "Substrait fixed-length binary type";
   let description = [{
     This type represents a substrait binary string of L bytes.
   }];
   let parameters = (ins "StringAttr":$value, "FixedBinaryType":$type);
+  let genVerifyDecl = 1;
   let assemblyFormat = "`<` custom<AddRemovePadding>($value, $type) `>`";
```
ingomueller-net, line 131: Unfortunately, I didn't see that yet in our discussion today: like for `FixedChar`, the length is now redundant in the assembly -- the string literal's length already carries that information. Should we not remove the explicit length from the assembly format here as well? (I think that the benefit of not needing the length by not allowing padding weighs more than the benefit of shortening strings with trailing null characters, so I think this was a win-win choice!)

Also, `AddRemovePadding` doesn't fit anymore.

ingomueller-net (approved):
LGTM overall, some minor comments.

test/Dialect/Substrait/literal.mlir
```diff
@@ -34,22 +34,22 @@
 // CHECK:      substrait.plan version 0 : 42 : 1 {
 // CHECK-NEXT:   relation
 // CHECK:         %[[V0:.*]] = named_table
-// CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : tuple<si1> -> tuple<si1, !substrait.var_char<6>> {
+// CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : tuple<si1> -> tuple<si1, !substrait.fixed_binary<10>> {
```
ingomueller-net, line 37: Maybe this is an artifact of the diff that I am looking at but this currently looks like it's going to *replace* a varchar test. That's not what we want, is it?

test/Dialect/Substrait/types.mlir
```diff
@@ -3,27 +3,41 @@
 
 // CHECK-LABEL: substrait.plan
 // CHECK:         relation
-// CHECK:         %[[V0:.*]] = named_table @t1 as ["a"] : tuple<!substrait.var_char<6>>
-// CHECK-NEXT:    yield %0 : tuple<!substrait.var_char<6>>
+// CHECK:         %[[V0:.*]] = named_table @t1 as ["a"] : tuple<!substrait.decimal<12, 2>>
+// CHECK-NEXT:    yield %0 : tuple<!substrait.decimal<12, 2>>
```
ingomueller-net, line 7: There are also some varchar to decimal changes here. Intentional?

test/Target/SubstraitPB/Export/literal.mlir
```diff
@@ -50,21 +50,17 @@ substrait.plan version 0 : 42 : 1 {
...
-// CHECK-NEXT:       }
+// CHECK-NEXT:            fixed_binary: "8181818181"
 
 substrait.plan version 0 : 42 : 1 {
   relation {
     %0 = named_table @t1 as ["a"] : tuple<si1>
-    %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.var_char<6>> {
-    ^bb0(%arg0: tuple<si1>):
-      %2 = literal #substrait.var_char<"hello", 6>
-      yield %2 : !substrait.var_char<6>
```
ingomueller-net, line 65: Type *changes* again. Intentional?

test/Target/SubstraitPB/Import/literal.textpb
```diff
@@ -71,12 +71,12 @@ version {
 # CHECK:      substrait.plan version 0 : 42 : 1 {
 # CHECK-NEXT:   relation
 # CHECK:         %[[V0:.*]] = named_table
-# CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : tuple<si1> -> tuple<si1, !substrait.var_char<5>> {
+# CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : tuple<si1> -> tuple<si1, !substrait.fixed_binary<10>> {
```
ingomueller-net, line 74: And again...

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -358,6 +369,35 @@ void printDecimalNumber(AsmPrinter &printer, DecimalType type,
   printer << "P = " << type.getPrecision() << ", S = " << type.getScale();
 }
 
+ParseResult parseFixedBinaryTypeRemoveLength(AsmParser &parser,
```
ingomueller-net, line 372: Maybe just `FixedBinary` or `FixedBinaryLiteral` is enough? Then we have "a custom assembly for the fixed binary literal" and "`(print|parse)FixedBinaryLiteral` prints|parses the fixed binary literal"?

ingomueller-net (approved):
LGTM! One minor comment left, plus a few warnings about potential rebase breakages. Merge away when done! 🚀

dshaaban01:
All comments addressed! I think it's ready 🚀

test/Target/SubstraitPB/Export/literal.mlir
dshaaban01: @ingomueller-net there was an extra copy of var_char tests (I think rebase mistakes) - so I deleted them.

dshaaban01:
Can I 🚀 ? Changed name + the other comments were rebase breakages that I fixed (extra set of varchar tests creeped into some of the test files)

PR #87, opened 2025-02-13
chore: change from pybind to nanobind

ingomueller-net:
Use nanobind instead of pybind11 for our Python bindings to follow what upstream MLIR is doing. This does not implement stub generation (i.e., the generation of `*.pyi` files), which didn't work nicely out of the box.


python/SubstraitDialects.cpp
```diff
@@ -7,19 +7,23 @@
...
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Signals.h"
 
-#include <vector>
+// Suppress warning from nanobind, otherwise CI with `-Werror` fails.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wcast-qual"
+#pragma clang diagnostic ignored "-Wnested-anon-types"
+#pragma clang diagnostic ignored "-Wzero-length-array"
+#include <nanobind/nanobind.h>
```
jpienaar, line 21: If you include mlir/include/mlir/Bindings/Python/Nanobind.h instead, I've already done similar squashing so that you don't have to think here.

jpienaar (approved):
LG thanks

python/SubstraitDialects.cpp
ingomueller-net: Yeah, nice, that's better! Thanks for the tip :)

PR #91, opened 2025-02-20
build: fix CMake dependencies for generated headers

ingomueller-net:
For a while, CI sporadically fails due to some generated headers not being found. I have finally understood why this happens. This PR fixes the dependencies among the generated headers and the main dialect target.


ingomueller-net:
@jpienaar, @mortbopet: it'd be great if one of you could stamp this PR; it's making our CI flaky.

mortbopet (approved):
✅

PR #92, opened 2025-02-21
chore: update `substrait-cpp` submodule to latest version (904a493fc)

mortbopet:
The currently tagged version of substrait-cpp is ~4 months old. Most notably, the current tag is missing some stuff that's required for it to be able to build on Windows - which in turn, is a prerequisite for getting `substrait-mlir` to build on Windows.


ingomueller-net (approved):
Awesome, thanks a lot for the work, here and in `substrait-cpp`!

ingomueller-net:
@mortbopet: I gather that you are able to compile `substrait-mlir` on Windows now? Let me know if you are missing anything else or have any input on future development.

mortbopet:
@ingomueller-net I'm unfortunately still fighting with it. Hitting the infamous, and good old, MSVC path length limit (separate from the general windows path-length limit, which _has_ been fixed).

There's some unfortunate interactions between dependencies-of-dependencies and LLVM build nesting which triggers this.

E.g., say that one has both `substrait-mlir` and `llvm` as a dependency of a 3rd project.
THen, one would typically, on windows, get into situations like this:

> C:/Users/${user}/${project}/_deps/llvm-build/tools/substrait_mlir/third_party/substrait-cpp/src/substrait/textplan/parser/grammar/antlr4_runtime/src/antlr4_runtime/runtime/Cpp/CMakeFiles/CMakeScratch/TryCompile-s3o9if/CMakeFiles/cmTC_945a8.dir/./

Which then, when generating files to that directory, exceeds the 260 character limit.

A bandaid could be to remove the `src/substrait/textplan/parser/grammar` part, which seems unnecessary (and AFAICT is just an artifact of _where_ in the `substrait-cpp` hierarchy the antlr build is being defined, which could easily be moved closer to the top level)... Will investigate...

PR #98, opened 2025-02-26
feat: add support for decimal values

mortbopet:
Adds the required type and attribute to support decimal values. Decimal values are represented as `TypedStrAttr`'d constrained strings with `DecimalType`s. Decimal types follow the constraints outlined in the [substrait docs](https://substrait.io/types/type_classes/).

### DIFF
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -176,6 +176,30 @@ def Substrait_VersionAttr : Substrait_Attr<"Version", "version"> {
   ];
 }
 
+def Substrait_DecimalAttr : Substrait_StaticallyTypedAttr<"Decimal", "decimal",
+                                      "DecimalType"> {
+  let summary = "Substrait decimal type";
+  let description = [{
+    This type represents a substrait decimal attribute type. The attribute is backed
+    by an IntegerAttr `value` which stores the 128-bit decimal value as an APInt.
+  }];
+  let parameters = (ins "DecimalType":$type, "IntegerAttr":$value);
+  let genVerifyDecl = 1;
+
+  // #substrait.decimal<"{integer}.{fraction}" : !substrait.decimal<{P}, {S}>>
+  let hasCustomAssemblyFormat = 1;
+
+  let builders = [
+    // Builder that accepts a string representation of the decimal value.
+    AttrBuilder<(ins "DecimalType":$type, "StringRef":$value)>,
+  ];
+
+  let extraClassDeclaration = [{
+    // Returns a string representation of the decimal value.
+    std::string decimalStr() const;
+  }];
+}
+
 //===----------------------------------------------------------------------===//
 // Helpers and constraints
 //===----------------------------------------------------------------------===//
@@ -193,7 +217,7 @@ def Substrait_AtomicAttributes {
     F64Attr, // FP64
     TypedStrAttr<Substrait_StringType>, // String
     TypedStrAttr<Substrait_BinaryType>, // Binary
-    TypedStrAttr<Substrait_DecimalType>, // Decimal
+    Substrait_DecimalAttr, // Decimal
     Substrait_TimestampAttr, // Timestamp
     Substrait_TimestampTzAttr, // TimestampTZ
     Substrait_DateAttr, // Date
--- a/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
+++ b/include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
@@ -193,6 +193,7 @@ def Substrait_AtomicAttributes {
     F64Attr, // FP64
     TypedStrAttr<Substrait_StringType>, // String
     TypedStrAttr<Substrait_BinaryType>, // Binary
+    TypedStrAttr<Substrait_DecimalType>, // Decimal
     Substrait_TimestampAttr, // Timestamp
     Substrait_TimestampTzAttr, // TimestampTZ
     Substrait_DateAttr, // Date
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -92,6 +92,25 @@ LogicalResult mlir::substrait::IntervalDaySecondAttr::verify(
 }
 
 //===----------------------------------------------------------------------===//
+// Substrait types
+//===----------------------------------------------------------------------===//
+
+LogicalResult mlir::substrait::DecimalType::verify(
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError, int32_t precision,
+    int32_t scale) {
+  if (precision > 38)
+    return emitError() << "precision must be in a range of [0..38] but got "
+                       << precision;
+
+  if (scale < 0 || scale > precision)
+    return emitError() << "scale must be in a range of [0..P] (P = "
+                       << precision << ")"
+                       << " but got " << scale;
+
+  return success();
+}
+
+//===----------------------------------------------------------------------===//
 // Substrait enums
 //===----------------------------------------------------------------------===//
 
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -13,6 +13,7 @@
 #include "mlir/Support/LogicalResult.h"
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/TypeSwitch.h" // IWYU pragma: keep
+#include "llvm/Support/Regex.h"
 
 using namespace mlir;
 using namespace mlir::substrait;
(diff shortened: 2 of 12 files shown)


ingomueller-net:
Resolves #96.

@mortbopet: Let me know if/when I should start reviewing (ideally using a GH request).

mortbopet:
@ingomueller-net feel free - i can't add you as a reviewer since I'm not part of the organization.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -91,6 +91,25 @@ LogicalResult mlir::substrait::IntervalDaySecondAttr::verify(
...
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError, int32_t precision,
+    int32_t scale) {
+  if (precision > 38)
+    return emitError() << "precision must be in a range of [0..38] but got "
+                       << precision;
+
+  if (scale < 0 || scale > precision)
+    return emitError() << "scale must be in a range of [0..P] (P = "
+                       << precision << ")"
+                       << " but got " << scale;
```
ingomueller-net, line 108: Ultra-nit:
```suggestion
                       << precision << ") but got " << scale;
```

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -675,6 +680,14 @@ importLiteral(ImplicitLocOpBuilder builder,
     auto attr = UUIDAttr::get(context, integer_attr);
     return builder.create<LiteralOp>(attr);
   }
+  case Expression::Literal::LiteralTypeCase::kDecimal: {
+    auto attr =
+        StringAttr::get(message.decimal().value(),
+                        DecimalType::get(context, message.decimal().precision(),
+                                         message.decimal().scale()));
```
ingomueller-net, line 687: Nit: factor out the type into a variable to make each expression shorter.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -193,6 +193,7 @@ def Substrait_AtomicAttributes {
     F64Attr, // FP64
     TypedStrAttr<Substrait_StringType>, // String
     TypedStrAttr<Substrait_BinaryType>, // Binary
+    TypedStrAttr<Substrait_DecimalType>, // Decimal
```
ingomueller-net, line 196: I am wondering if this is the right attribute type and I think that a different choice would have some advantages: the spec says that decimal values represent 16-bytes, so we can make this an `i128`. This would immediately ensure that we never have literal values that are (1) not numbers and (2) out of bound, which the current design does not ensure. Using integers instead of strings might also be slightly faster.

The downside of this might seem that we don't have the point in the assembly representation. But that can be easily changed with a custom printer/parser that takes `S` into account. I'd be happy if that weren't implemented and just hinted at with a `TODO`.

Finally, the current design also doesn't check if the number fits into the limit imposed by `P`.

What I envision to solve all of this is to create new attribute in our dialect that has one `i128` parameter, verifies that it fits into the limit imposed by `P`, and (eventually) gets a custom printer/parser.

@dshaaban01 and I are still discussing to do something similar for `fixedchar<L>` in #81 but haven't finished yet, so we don't have much to copy from, yet.

ingomueller-net (commented):
@mortbopet: awesome, so happy to see your contribution :) This looks good overall but I have a suggestion with respect to how to model the attribute type. WDYT?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
mortbopet: Having a decimal attribute that's backed by an i128 was also my initial thought. I ended up going down this route mainly out of convenience, given how easily it fit into the `TypedStrAttr` idiom that's already used.

But principally, i agree with you. Let me see if i can get time this week to refactor this PR. If not, it's also an option to merge as-is with a follow-up task to modify the representation.

ingomueller-net: Yeah, I guess that's an option. It's true that we don't have any other paramtrized type, yet, so there isn't any other pattern to follow. If you don't want to work on one, you can merge with the current design and let me have a stab at it.

mortbopet:
@ingomueller-net pushed a refactored version of how to represent the attribute.
Now, a new `DecimalAttr` has been introduced, of type `DecimalType`, backed by an `IntegerAttr` (verified as an `APInt` of 128 bits). It's represented by a string, that is en/decoded as per the substrait spec. This also means that `S` and `P` are now validated.

test/Dialect/Substrait/literal.mlir
```diff
@@ -227,3 +227,18 @@ substrait.plan version 0 : 42 : 1 {
...
+// CHECK:         %[[VAL_10:.*]] = substrait.literal #substrait.decimal<"0.123" : !substrait.decimal<3, 3>>
+// CHECK:         %[[VAL_11:.*]] = substrait.literal #substrait.decimal<"123.0" : !substrait.decimal<3, 0>>
+// CHECK:         %[[VAL_12:.*]] = substrait.literal #substrait.decimal<"123.0" : !substrait.decimal<3, 0>>
+// CHECK:         %[[VAL_13:.*]] = substrait.literal #substrait.decimal<"1111111111111111111111111111111111113.2" : !substrait.decimal<38, 1>>
+%d0 = substrait.literal #substrait.decimal<"1234567.89" : !substrait.decimal<9, 2>>
+%d1 = substrait.literal #substrait.decimal<"0.123" : !substrait.decimal<3, 3>>
+%d2 = substrait.literal #substrait.decimal<"000.123" : !substrait.decimal<3, 3>>
+%d3 = substrait.literal #substrait.decimal<"123.0" : !substrait.decimal<3, 0>>
+%d4 = substrait.literal #substrait.decimal<"123.000" : !substrait.decimal<3, 0>>
+%d5 = substrait.literal #substrait.decimal<"1111111111111111111111111111111111113.2" : !substrait.decimal<38, 1>>
```
ingomueller-net, line 244: Nice! This is so much simpler than the whole plans that we have used until now!

mortbopet:
> Nice! This now verifies all constraints. I didn't expect it to be so much effort but I can't come up with suggestions for simplification. I must admit that I underestimated that a bit. Thanks for the effort!

I agree - a bit more than i thought. It's a nice combination of having no built-in support for `i128` in MLIR, non-trivial decimal string printing/parsing, and the discrepancy between the underlying storage and the displayed value.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -91,6 +92,151 @@ LogicalResult mlir::substrait::IntervalDaySecondAttr::verify(
...
+                                           mlir::StringRef(value));
+}
+
+void DecimalAttr::print(AsmPrinter &odsPrinter) const {
+  Builder odsBuilder(getContext());
+  odsPrinter << "<";
+  odsPrinter << "\"" << decimalStr() << "\"";
+  odsPrinter << " : ";
+  odsPrinter.printType(getType());
+  odsPrinter << ">";
```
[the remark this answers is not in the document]
mortbopet: It's unfortunately a bit hard to do away with the string here. An `int|.|int` sequence can't be done with the built-in parser utilities due to tokenization (we're don;t want whitespace in between the values).  
Then there's the option of keyword parsing - but, the built-in parser doesn't allow keywords to be numerically prefixed.  
And finally, we don't have access to the internal token state of the parser - which would allow us to do something like:
```cpp
  StringRef value = odsParser.getTokenSpelling();
  parser.consumeToken();
```

Any other ideas?

test/Dialect/Substrait/literal.mlir
mortbopet: Yea, i was wondering why this file wasn't just using the literals directly instead of nesting them inside a relation... perhaps this file was initially written when `substrait.literal` had stricter parent op requirements?

test/Dialect/Substrait/literal.mlir
```diff
@@ -227,3 +227,18 @@ substrait.plan version 0 : 42 : 1 {
     yield %1 : tuple<si1, si1, si8, si16, si32, si64>
   }
 }
+
+// -----
+
+// CHECK:         %[[VAL_8:.*]] = substrait.literal #substrait.decimal<"1234567.89" : !substrait.decimal<9, 2>>
```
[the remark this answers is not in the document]
mortbopet: Changed; i personally just tend to do whatever `generate-test-checks.py` throws at me - less work :).

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -91,6 +92,151 @@ LogicalResult mlir::substrait::IntervalDaySecondAttr::verify(
...
+  // Trim trailing zeros.
+  size_t lastNonZero = value.find_last_not_of('0');
+  if (lastNonZero == StringRef::npos)
+    lastNonZero = value.size() - 1;
+  value = value.substr(0, lastNonZero + 1);
+
+  // Verify scale.
+  size_t decimalPos = value.find('.');
+  size_t detectedScale = value.size() - decimalPos - 1;
+  if (detectedScale != type.getScale()) {
```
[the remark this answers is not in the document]
mortbopet: Good catch; `>` is indeed sufficient. And as you hint to, be do need an additional step to shift-in trailing zeros in case `detectedScale != actualScale`.

mortbopet:
> Feel free to pick this up now. Otherwise, I might look more into it tomorrow.

I think i'll leave this branch as is (because i need to get back to some internal stuff 🙈), but go ahead and push to it if you do find that we can represent the full decimal range without unintentional rounding and truncation.

ingomueller-net (commented):
As commented there, I created #108 with a follow-up.

ingomueller-net:
Closing as this PR was the base for #108, which I just merged.

PR #99, opened 2025-02-27
build: use and adapt to more recent Protobuf from `substrait-cpp`

ingomueller-net:
This PR upgrades the codebase to the more recent Protobuf version downloaded and compiled by `substrait-cpp`. Previously, we used `find_package` from our CMake files, even though we don't use those targets directly and only use those imported by `substrait-cpp` already. Also, since recently, `substrait-cpp` can download Protobuf if none is installed in the systems, so we rely on that mechanism in CI now. Since that version is newer than the one we used in CI until now, the PR also updates the code base to the new version.

The PR suppresses some warnings in the headers of the new Protobuf version, which are errors in CI due to our using of `-Werror`, by setting some diagnostic pragmas. However, it would be better if those headers were imported as `SYSTEM` headers by `substrait-cpp`, but I still need to figure out how to do that and update that repository accordingly.

### DIFF
--- a/lib/Target/SubstraitPB/Export.cpp
+++ b/lib/Target/SubstraitPB/Export.cpp
@@ -18,12 +18,17 @@
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/ErrorHandling.h"
 
+// TODO(ingomueller): Find a way to make `substrait-cpp` declare these headers
+// as system headers and remove the diagnostic fiddling here.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Weverything"
 #include <google/protobuf/text_format.h>
 #include <google/protobuf/util/json_util.h>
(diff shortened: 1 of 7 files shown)


lib/Target/SubstraitPB/Export.cpp
```diff
@@ -18,23 +18,28 @@
...
 using namespace mlir::substrait::protobuf_utils;
 using namespace ::substrait;
 using namespace ::substrait::proto;
 
-namespace pb = google::protobuf;
-
 namespace {
 
+namespace pb = ::google::protobuf;
+
```
mortbopet, line 42: This nesting of namespaces might also have solved some windows issues... We'll see later today!

mortbopet (approved):

ingomueller-net:
I'll only be able to remove the diagnostic shenanigans once llvm/llvm-project#129254 has be resolved.

PR #101, opened 2025-02-28
feat: collection of MSVC/Windows-related fixes

mortbopet:
This PR adds some minor fixes that MSVC complains about on Windows.

* fix pb namespace redefinition: namespace `pb` may already be defined [here](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/extension_set.h#L77) (in case the depending build gets protobuf from there).
* Modifies usage of `google::protobuf::util::Status` to `auto`. `google::protobuf::util::Status` alias to `absl::Status` but that isn't being handled properly on MSVC. `auto` should make all platforms happy :).
* Added missing link to `LLVMSupport` for the `_substraitDialects` Python library.


mortbopet:
dropping - most of these changes were due to mismatched protobuf versions.

PR #105, opened 2025-03-04
feat: windows nits and CI

mortbopet:
Collection of a few nits to make MSVC happy + CI support for Windows.

Python binding tests have been XFAIL'ed for now due to .dll initialization issues for Windows. This is most likely due to some erroneous configuration re. python path or perhaps incompatible pybound(nanobound) modules are being picked up. Since I won't be needing those bindings anytime soon, i suggest getting the PR in as is (getting Windows building + CI) and creating an issue to track the python issue.

### DIFF
--- a/include/substrait-mlir/Target/SubstraitPB/Export.h
+++ b/include/substrait-mlir/Target/SubstraitPB/Export.h
@@ -21,7 +21,7 @@ using LogicalResult = llvm::LogicalResult;
 
 namespace substrait {
 
-LogicalResult
+mlir::LogicalResult
 translateSubstraitToProtobuf(Operation *op, llvm::raw_ostream &output,
                              substrait::ImportExportOptions options = {});
 
--- a/include/substrait-mlir/Target/SubstraitPB/Export.h
+++ b/include/substrait-mlir/Target/SubstraitPB/Export.h
@@ -17,13 +17,12 @@ class LogicalResult;
 }
 namespace mlir {
 class Operation;
-using LogicalResult = llvm::LogicalResult;
 
 namespace substrait {
 
-mlir::LogicalResult
+llvm::LogicalResult
 translateSubstraitToProtobuf(Operation *op, llvm::raw_ostream &output,
-                             substrait::ImportExportOptions options = {});
+                             mlir::substrait::ImportExportOptions options = {});
 
 } // namespace substrait
 } // namespace mlir
--- a/include/substrait-mlir/Target/SubstraitPB/Export.h
+++ b/include/substrait-mlir/Target/SubstraitPB/Export.h
@@ -13,16 +13,17 @@
 #include "llvm/Support/raw_ostream.h"
 
 namespace llvm {
-class LogicalResult;
+struct LogicalResult;
 }
 namespace mlir {
 class Operation;
+using LogicalResult = llvm::LogicalResult;
 
 namespace substrait {
 
-llvm::LogicalResult
-translateSubstraitToProtobuf(Operation *op, llvm::raw_ostream &output,
-                             mlir::substrait::ImportExportOptions options = {});
+LogicalResult translateSubstraitToProtobuf(Operation *op,
+                                           llvm::raw_ostream &output,
+                                           ImportExportOptions options = {});
 
 } // namespace substrait
 } // namespace mlir
--- a/lib/Target/SubstraitPB/Export.cpp
+++ b/lib/Target/SubstraitPB/Export.cpp
@@ -1501,7 +1501,7 @@ SubstraitExporter::exportOperation(Operation *op) {
 
 } // namespace
 
-LogicalResult mlir::substrait::translateSubstraitToProtobuf(
+mlir::LogicalResult mlir::substrait::translateSubstraitToProtobuf(
     Operation *op, llvm::raw_ostream &output,
     substrait::ImportExportOptions options) {
   SubstraitExporter exporter;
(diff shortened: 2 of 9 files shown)


mortbopet:
@ingomueller-net after much mucking about, this is finally passing and ready for review. Windows build times is (probably as expected) not as fast as Ubuntu, but not horrible either (13m); IMO still within the realms of keeping it as a PR gate.

include/substrait-mlir/Target/SubstraitPB/Export.h
```diff
@@ -17,13 +17,12 @@ class LogicalResult;
...
 class Operation;
-using LogicalResult = llvm::LogicalResult;
 
 namespace substrait {
 
-LogicalResult
+llvm::LogicalResult
 translateSubstraitToProtobuf(Operation *op, llvm::raw_ostream &output,
-                             substrait::ImportExportOptions options = {});
+                             mlir::substrait::ImportExportOptions options = {});
```
ingomueller-net, line 25: Out of curiosity and to be able to predict what I have to do for new code: Why is this change necessary? Could we also remove both namespaces, i.e., use just `ImportExportOptions`?

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -1501,23 +1501,19 @@ SubstraitExporter::exportOperation(Operation *op) {
 
 } // namespace
 
-namespace mlir {
-namespace substrait {
-
-namespace pb = ::google::protobuf;
```
ingomueller-net, line 1507: Same intention as above: why does this not work?

ingomueller-net (approved):
LGTM! Thanks for this heroic effort! I know that debugging CI can be a pain if you only get a feedback from your change 15 minutes later (or sometimes hours!)...

include/substrait-mlir/Target/SubstraitPB/Export.h
mortbopet: `mlir::substrait::ImportExportOptions` could indeed be set to just `ImportExportOptions`.

and... Hah! Typing up an explanation for the `LogicalResult` issue, i actually stumbled upon the solution (which has tricked me before.

User:
```sh
> dumpbin.exe /SYMBOLS Dialects.cpp.obj
0E1 00000000 UNDEF  notype ()    External     | ?translateSubstraitToProtobuf@substrait@mlir@@YA?AULogicalResult@llvm@@PEAVOperation@2@AEAVraw_ostream@4@UImportExportOptions@12@@Z (struct llvm::LogicalResult __cdecl mlir::substrait::translateSubstraitToProtobuf(class mlir::Operation *,class llvm::raw_ostream &,struct mlir::substrait::ImportExportOptions))
```

Definition:
```sh
> dumpbin.exe /SYMBOLS Export.cpp.obj
675 00000000 SECT313 notype ()    External    | ?translateSubstraitToProtobuf@substrait@mlir@@YA?AVLogicalResult@llvm@@PEAVOperation@2@AEAVraw_ostream@4@UImportExportOptions@12@@Z (class llvm::LogicalResult __cdecl mlir::substrait::translateSubstraitToProtobuf(class mlir::Operation *,class llvm::raw_ostream &,struct mlir::substrait::ImportExportOptions))
```

The difference being that MSVC is picky about whether or not something is a `struct` or a `class`, since it's included in the name mangling of a symbol. Hence, `Export.h` pre-declared `LogicalResult` as a `class` where it in fact is defined as a `struct`.

lib/Target/SubstraitPB/Export.cpp
mortbopet: For this, the issue is slightly different than the above.

MSVC is generally bad at not having fully qualified free function definitions when they're declared inside a nested namespace, e.g.:
```cpp
// .h
namespace foo {
void bar();
}

// .cpp

// Problematic on MSVC
namespace {
void bar() {}
}

// Always works on MSVC
void foo::bar(){}
```

mortbopet:
Ohh yes, it took a while. Before `ccache` kicked in (which only started once i realized that `ccache` wasn't going to work and `sccache` is what one should use), it was 2 hour iteration times 🫠.

include/substrait-mlir/Target/SubstraitPB/Export.h
ingomueller-net: Uh, that struct vs class thing made problems earlier... But happy you found the root cause :)

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: Oh, I realize that there are two issues to discuss: (1) Definitions in namespaces, which I understand MSVC has problems with. (BTW, it should say `namespace foo {` underneath the comment `// Problematic on MSVC`, right?) And (2) the namespace alias (`using pb = ::google::protobuf`). That's handy because it removes some verbosity, no? BTW, why isn't that a problem for `using nb = nanobind` in `SubstraitDialects.cpp`?

mortbopet: The reason for moving `using pb = ::google::protobuf` in the in those files is due to a namespace redeclaration conflict (at least on MSVC) in the root namespace between that declaration, and [this](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/extension_set.h#L77) in protobuf. To me it then seems like a necessary evil, since the free functions (on MSVC) needs to be fully qualified (as described above), so we don't _have_ a namespace to place that `using pb = ...` into.

Which also means that the `using nb = nanobind` should be perfectly fine in `SubstraitDialects.cpp` since it doesn't conflict with anything.

PR #107, opened 2025-03-10
feat: mark LiteralOp as ConstantLike

mortbopet:
... and add folder which simply returns the underlying attribute.


ingomueller-net (approved):
Yep, makes total sense. Thanks! LGTM!

We could still add a test to `test/Dialect/Substrait/canonicalize.mlir`.

ingomueller-net:
> We could still add a test to `test/Dialect/Substrait/canonicalize.mlir`.

On a second thought: I am not even sure where this applies currently. Do you have a situation in mind?

PR #109, opened 2025-03-13
test: add tooling for new `Round-Trip` test type

ingomueller-net:
This PR adds some infrastructure for testing that Substrait plans round-trip through `substrait-mlir` without loosing any information. This is currently based on the JSON format and consist in comparing (1) a normalized version of the input plan with (2) a normalized version of the plan after round-tripping it through `substrait-mlir`. JSON is maybe the format that makes this easiest and most robust since normalizing JSON can be done with mature tools (deserialize from and reserialize to JSON). The PR also extends LIT to run these tests and adds a test case that exercises the round-trip testing infrastructure on a couple of plans based on the `plan.mlir` cases from the current dialect tests.

### DIFF
--- a/requirements-unfrozen.txt
+++ b/requirements-unfrozen.txt
@@ -7,6 +7,7 @@ duckdb==1.1.3
 ibis==3.3.0
 ibis-framework==8.0.0
 ibis-substrait==3.2.0
+json5
 pyarrow
 pyright
 yapf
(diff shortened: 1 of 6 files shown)


ingomueller-net:
OK, CI passes under Windows. However, after understanding your idea with the `--verify-roundtrip` test, I wonder whether that isn't quite a bit more elegant than this...

ingomueller-net:
Thinking about this again: I guess it's better to take this ~as is and concentrate on increasing coverage instead. We can still address #109 later and I can add links to it in this PR.

mortbopet (approved):

PR #117, opened 2025-03-26
chore: clean up + alphabetical reordering of decimal and add missing decimal test

dshaaban01:
Reorders decimal type + attr to be in alphabetical order, reorders tests to fit current ordering, adds missing decimal type test.


ingomueller-net (approved):
LGTM! Thanks a lot for the clean-up! (Note that something most of gone wrong in the literal test.)

PR #119, opened 2025-03-31
feat: add `advanced_extension` attribute to remaining `*Rel`s

ingomueller-net:
This PR adds the `advanced_extension` to the remaining ops corresponding to `*Rel` message types by adding the corresponding attribute, extending the assembly format, and adding custom builders to be able to skip the new attribute.

### DIFF
--- a/test/Target/SubstraitPB/Export/aggregate.mlir
+++ b/test/Target/SubstraitPB/Export/aggregate.mlir
@@ -243,7 +243,6 @@ substrait.plan version 0 : 42 : 1 {
   }
 }
 
-
 // -----
 
 // Check op with advanced extension.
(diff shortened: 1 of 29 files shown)


test/Target/SubstraitPB/Export/aggregate.mlir
```diff
@@ -242,3 +242,35 @@ substrait.plan version 0 : 42 : 1 {
     yield %1 : tuple<si32, si32, si32, si32, si32, si32, si32>
   }
 }
+
```
dshaaban01, line 245: Nit: extra line

dshaaban01 (approved):

ingomueller-net:
> LGTM! I've previously had to sanitize some input for these fields, so great that that step can be removed now :).

I am glad this is useful to somebody 😉 TBH, I mostly implemented this because it brings our coverage of five or so `Rel`s to 100%, which gives me the sense to complete stuff.

OOC, what did you have to sanitize? #118 fixed the values used for `google.protobuf.Any` messages; maybe that solved your problem?

PR #121, opened 2025-04-04
introduce `ListType<T>` type and attribute

dshaaban01:
This PR introduces the container type `ListType<T>` and `ListAttr<T>` which represent values of type T, where T can be an atomic or parameterized type.


include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -290,10 +290,28 @@ def Substrait_ParametrizedAttributes {
 /// Attribute of one of the currently supported atomic types.
 def Substrait_AtomicAttribute : AnyAttrOf<Substrait_SimpleAttributes.attrs#Substrait_ParametrizedAttributes.attrs>;
 
+def Substrait_ListAttr : Substrait_Attr<"List", "list", [TypedAttrInterface]> {
+  let summary = "Substrait list attribute";
+  let description = [{
+    This attribute represents a list of atomic attributes, parameterized with a ListType.
```
ingomueller-net, line 296: Is the restriction to atomic element types intentional? If so, we should probably add a TODO (here and to the type if it has the same restriction). Note that we can have "lists of lists" and "structs of lists of structs of lists" etc. in Substrait.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -180,14 +180,32 @@ def Substrait_AtomicTypes {
 /// One of the currently supported atomic types.
 def Substrait_AtomicType :
   AnyTypeOf<Substrait_SimpleTypes.types#Substrait_ParametrizedTypes.types>;
+  
+def Substrait_ListType : Substrait_Type<"List", "list"> {
+  let summary = "Substrait list type";
+  let description = [{
+    This type represents a list type.
+  }];
+  let parameters = (ins Substrait_AtomicType:$type);
```
ingomueller-net, line 189: Same here: if the restriction to atomic types is intentional, then add a TODO. Adding support for arbitrary nesting of lists and structs is probably not trivial and it may be a good strategy to do it in two steps.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -116,6 +116,28 @@ LogicalResult mlir::substrait::IntervalDaySecondAttr::verify(
...
+    ListType listType) {
+
+  mlir::Type expectedType =
+      listType.getType(); // The atomic type the list is parameterized with
+
+  for (mlir::Attribute attr : value) {
+    auto typedAttr = mlir::dyn_cast<mlir::TypedAttr>(attr);
+    if (!typedAttr) {
+      return emitError()
+             << "ListAttr values must be typed attributes, but got: " << attr;
```
ingomueller-net, line 130: Nit: single-quoted class name.
```suggestion
             << "'ListAttr' values must be typed attributes, but got: " << attr;
```

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -116,6 +116,28 @@ LogicalResult mlir::substrait::IntervalDaySecondAttr::verify(
...
+    }
+
+    if (typedAttr.getType() != expectedType) {
+      return emitError() << "Mismatched element type in ListAttr: expected "
+                         << expectedType << ", but got " << typedAttr.getType();
+    }
+  }
```
ingomueller-net, line 137: Ultra-nit: blank line after the loop body.
```suggestion
  }

```

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -144,6 +166,24 @@ LogicalResult mlir::substrait::DecimalType::verify(
...
+  if (!mlir::isa<mlir::IntegerType, mlir::Float32Type, mlir::Float64Type,
+                 substrait::StringType, substrait::BinaryType,
+                 substrait::TimestampType, substrait::TimestampTzType,
+                 substrait::DateType, substrait::TimeType,
+                 substrait::IntervalYearMonthType,
+                 substrait::IntervalDaySecondType, substrait::UUIDType,
+                 substrait::FixedCharType, substrait::VarCharType,
+                 substrait::FixedBinaryType, substrait::DecimalType>(type))
+    return emitError()
+           << "ListType must be parameterized with an atomic type, but got: "
```
ingomueller-net, line 181: Nit: single-quoted class name.
```suggestion
           << "'ListType' must be parameterized with an atomic type, but got: "
```

include/substrait-mlir/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -180,14 +180,32 @@ def Substrait_AtomicTypes {
...
+  let assemblyFormat = [{ `<` $type `>` }];
+  let genVerifyDecl = 1;
```
ingomueller-net, line 191: I think it is possible and most certainly preferrable to verify this using tablegen (more). Take a look at `TupleOf` in `CommonTypeConstraints.td` upstream. (`NestedTupleOf` will then be helpful as inspiration for nested Substrait structs, lists, and maps...). One problem here is that we don't see in this file which types can be used, plus we now have to maintain *two* lists in two different files, which risk to run out of sync at some point.

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -929,17 +945,17 @@ SubstraitExporter::exportOperation(FilterOp op) {
...
+SubstraitExporter::exportAttribute(Attribute value) {
   // Build `Literal` message depending on type.
-  Attribute value = op.getValue();
   mlir::Type literalType = getAttrType(value);
   auto literal = std::make_unique<Expression::Literal>();
 
   // `IntegerType`s.
   if (auto intType = dyn_cast<IntegerType>(literalType)) {
     if (!intType.isSigned())
-      op->emitOpError("has integer value with unsupported signedness");
```
ingomueller-net, line 942: Why this change (here and two times below)?

Maybe the intention was to only emit an error message below in `exportOperation(LiteralOp)`. However, we have more information here and I think it's better to emit the more precise message here (and none below).

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -733,6 +740,27 @@ importLiteral(ImplicitLocOpBuilder builder,
     auto attr = DecimalAttr::get(context, type, value);
     return builder.create<LiteralOp>(attr);
   }
+  case Expression::Literal::LiteralTypeCase::kList: {
+    const Expression::Literal::List &listType = message.list();
+    llvm::SmallVector<Attribute> listElements;
+    listElements.reserve(listType.values_size());
+    for (const Expression_Literal &element : listType.values()) {
+      // TODO: Create importAttribute function to avoid creating redundant
```
ingomueller-net, line 748: Is this TODO still up-to-date?

test/Dialect/Substrait/literal.mlir
```diff
@@ -1,6 +1,37 @@
...
+// CHECK-NEXT:      yield %[[V2]] : !substrait.list<!substrait.fixed_binary<4>>
+// CHECK-NEXT:    }
+// CHECK-NEXT:    yield %[[V1]] : tuple<si1, !substrait.list<!substrait.fixed_binary<4>>>
+
+substrait.plan version 0 : 42 : 1 {
+  relation {
+    %0 = named_table @t1 as ["a"] : tuple<si1>
+    %1 = project %0 : tuple<si1> -> tuple<si1, !substrait.list<!substrait.fixed_binary<4>>> {
+    ^bb0(%arg : tuple<si1>):
+      %bytes = literal #substrait.list<[
```
ingomueller-net, line 23: Nit: variable name.
```suggestion
      %list = literal #substrait.list<[
```

test/Dialect/Substrait/literal.mlir
```diff
@@ -1,6 +1,37 @@
...
+                        #substrait.fixed_binary<"8181">,
+                        #substrait.fixed_binary<"8181">,
+                        #substrait.fixed_binary<"8181">], !substrait.list<!substrait.fixed_binary<4>>>
```
ingomueller-net, line 26: There is a lot of redundancy in the assembly format but I think it's hard to reduce. Unfortunately, there is no built-in way to remove the `#substrait.` and `!substrait.` prefixes, for example. In the LLVM dialect, there is a work-around that parses type names nested in other LLVM types manually, so the `!llvm.` prefix can be omitted there. We could do something like that. What would be even more concise is if we could do something like `#substrait.list<["8181", "8181", "8181"], list<fixed_binary<4>>>` and use the type for the parsing of the values (potentially swapping the order of the values and the type). Both are probably quite a lift, and that "only" for the assembly format, so this is a clear candidate for skipping now (and potentially forever).

ingomueller-net (commented):
Nice progress! A few comments, most of them minor and some optional.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
```diff
@@ -290,10 +290,28 @@ def Substrait_ParametrizedAttributes {
...
+  }];
+  let parameters = (ins "mlir::ArrayAttr":$value, "ListType":$type);
```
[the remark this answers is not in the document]
dshaaban01: Oh gosh, that was a total accident.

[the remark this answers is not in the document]
dshaaban01: Oh no wait: I'm pretty sure that `ListType` is the syntax I should use in this file, not `Substrait_ListType`. I did the same thing for my other attributes that referenced types, i.e. VarCharAttr has a parameter called `VarCharType`, NOT `Substrait_VarCharType`.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
dshaaban01: Will add TODO. I think better to do in a separate PR.

lib/Target/SubstraitPB/Export.cpp
dshaaban01: So I created a function called `exportAttribute(Attribute value)` that contains all of the logic that was previously in `exportOperation(LiteralOp)`.

Then I altered exportOperation(LiteralOp op) to look like this. 

```
FailureOr<std::unique_ptr<Expression>>
SubstraitExporter::exportOperation(LiteralOp op) {
  // Build `Literal` message depending on type.
  Attribute value = op.getValue();
  auto literal = exportAttribute(value);

  if (failed(literal))
    return op->emitOpError("has unsupported value");

  // Build `Expression` message.
  auto expression = std::make_unique<Expression>();
  expression->set_allocated_literal(literal->release());

  return expression;
}
```

Since `exportAttribute(Attribute value)` does not have access to the op, I can't emit an error message via the op. And not sure how to pass the error string into `failure()` failures. Wanted to talk to you about this.

lib/Target/SubstraitPB/Import.cpp
dshaaban01: This TODO is to implement the same logic in the import.cpp, as what I did in export.cpp (i.e. with the creation of `exportAttribute(Attribute value)` and augmentation of `exportOperation(LiteralOp)`.)

Did not want to implement it yet until we spoke about you're thoughts on my export.cpp implementation

test/Dialect/Substrait/literal.mlir
dshaaban01: Yea it's super redundant, also wasn't quite sure how to handle. :( Should I maybe add a TODO to implement a printer/parser?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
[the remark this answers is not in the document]
ingomueller-net: And the other syntax doesn't work? Maybe we've been doing it wrong the whole time? I don't have the details in my head and haven't checked, so maybe there *is* a reason why we are doing it this way...

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: Ah, I see. Three patterns come to mind: (1) Use `InFlightDiagnostics` but maybe you have the same problem of not having an op; (2) provide a `Location loc` argument and emit the error at that location; or (3) provide an `emitError` function argument and use that to emit the error.

test/Dialect/Substrait/literal.mlir
ingomueller-net: Yeah, TODO is definitely fine. As I said, all alternatives that come to my mind are significantly more complex to implement, which is a good reason not to do it, at least immediately.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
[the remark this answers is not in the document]
dshaaban01: https://mlir.llvm.org/docs/DefiningDialects/AttributesAndTypes/#parameters

`Substrait_ListType` is not `AttrParameter` or `TypeParameter`, but rather a `TypeDef`. Therefore, based on my understanding of what the above link says, we have to use the "raw c++Type string instead", which in this case would be `ListType`. 

(Other syntax does not work btw, I tried)

lib/Target/SubstraitPB/Export.cpp
dshaaban01: Okay will do. And do you agree with the restructuring? Should I similarly implement `exportAttribute(Attribute value)` ?

include/substrait-mlir/Dialect/Substrait/IR/SubstraitAttrs.td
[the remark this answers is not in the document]
ingomueller-net: OK, cool, thanks for confirming.

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: OK, cool. Yeah, I think it's a good idea to cut this mega function into pieces :) Thanks!

dshaaban01:
Note to self: Final TODO for this PR, implement `ListOf` (similar to [TupleOf](https://github.com/llvm/llvm-project/blob/120e940356561035bf37bc4fcb6fab771e7a9b7d/mlir/include/mlir/IR/CommonTypeConstraints.td#L4)).

PR #125, opened 2025-04-10
feat: dyn_cast_or_null EmitOp's input defining op

mortbopet:
... in case the defining op is a block argument, which in turn would previously result in a crash.


ingomueller-net (approved):
LGTM, thanks!

PR #126, opened 2025-04-15
feat: make plan body ops an interface

mortbopet:
Partly motivated by not having a long list of legal ops, mostly motivated by increasing the interoperability of the dialect, allowing other dialects to mark ops with `PlanBodyOpInterface`, making them legal to nest within a `Substrait_PlanOp`.


ingomueller-net (approved):
Nice! LGTM.

Issue #97, opened 2025-02-26
Implement `HashJoinRel`


ingomueller-net:
The only addition compare to `JoinRel` seems to be `message ComparisonJoinKey`. Maybe this could look like this in MLIR:

```
%1 = subtrait.hash_join %0, %1 on {
  ^bb0(%arg0: tuple<si32, si32>, %arg1: tuple<si32, si32, si32>):
  }

Trion129:
@ingomueller-net is it up for grabs or someone is already working on this? I am thinking of taking this up.

ingomueller-net:
From a [discussion](https://substrait.slack.com/archives/C08GGQVKR46/p1744646689440419) on Slack:

> Needed some thought from you regarding the HashJoinRel. To handle the ComparisonJoinKey message, I am currently keeping ExpressionType attributes for left and right key exprs which show in assemblyFormat as,
> ```
> 2 = hash_join inner %0, %1 on
>      left_keys %left_key : si32, 
>      right_keys %right_key : si32
>      : tuple<si32, si32>, tuple<si32, si32, si32> -> tuple<si32, si32, si32, si32, si32> {
>   ^bb0(%arg0: tuple<si32, si32>, %arg1: tuple<si32, si32, si32>):
>     %3 = field_reference %arg0[0] : tuple<si32, si32>
>     %4 = field_reference %arg1[0] : tuple<si32, si32, si32>
>     %5 = compare not_distinct_from %3, %4 : (si32, si32) -> si1
>     yield %5 : si1
> }
> ```
> Should I omit these? Then it becomes something like...
> ```
> %2 = hash_join inner %0, %1 on
>      : tuple<si32, si32>, tuple<si32, si32, si32> -> tuple<si32, si32, si32, si32, si32> {
>   ^bb0(%arg0: tuple<si32, si32>, %arg1: tuple<si32, si32, si32>):
>     %3 = field_reference %arg0[0] : tuple<si32, si32>
>     %4 = field_reference %arg1[0] : tuple<si32, si32, si32>
>     %5 = compare not_distinct_from %3, %4 : (si32, si32) -> si1
>     yield %5 : si1
> }
> ```
> Which MLIR design is preferred?

[...]

PR #129, opened 2025-04-28
feat: `HashJoinOp` mlir implementation

Trion129:
This commit adds MLIR implementation for HashJoinOp
This tries to follow a similar MLIR pattern as in issue https://github.com/substrait-io/substrait-mlir-contrib/issues/97

1. Substrait.cpp: Verifier implementations for HashJoin and KeyComparison operations. For Key comparisons tries to allow similar types. Might not be the best way.
2. Import.cpp: Converts HashJoinRel into Filter->HashJoin->ConditionRegion. Post join filter is lifted to filter, key comparison operator added to nested region.
3. Export.cpp: It will not be 1->1 mapping form import to export as currently it does not convert Filter->HashJoin into HashJoin, I have assumed it might not be needed and these 2 can convert to seperate relations.
4. *.td files: Added enum and Op implementations.

Skipped work:
Doesn't implement custom_function_id. It is not needed by us at the moment, so left a TODO for now.

### DIFF
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -1078,13 +1078,25 @@ HashJoinOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
 }
 
 LogicalResult KeyComparisonOp::verify() {
-  Type lhsType = getLhs().getType();
-  Type rhsType = getRhs().getType();
-
-  if (lhsType != rhsType) {
-    return emitOpError() << "operands must have the same type but got "
-                         << lhsType << " and " << rhsType;
-  }
+  auto &res =
+      llvm::TypeSwitch<mlir::Type, LogicalResult>(getLhs().getType())
+          .Case<substrait::VarCharType, substrait::StringType>([&](auto) {
+            if (!mlir::isa<substrait::VarCharType, substrait::StringType>(
+                    getRhs()))
+              return this->emitError("Invalid rhs type for string comparison, "
+                                     "expected string-like type but got")
+                     << getRhs().getType();
+            return success();
+          })
+          .Case<IntegerType>([&](auto) {
+            if (!mlir::isa<IntegerType, substrait::DecimalType>(getRhs()))
+              return this->emitError("Invalid rhs type for integer comparison, "
+                                     "expected integer-like type but got")
+                     << getRhs().getType();
+            return success();
+          });
+  if (failed(res))
+    return res;
 
   return success();
 }
--- a/lib/Target/SubstraitPB/Import.cpp
+++ b/lib/Target/SubstraitPB/Import.cpp
@@ -675,8 +675,6 @@ wrapFilterOnJoin(ImplicitLocOpBuilder builder, RelOpInterface hashJoin,
 static mlir::FailureOr<RelOpInterface>
 importHashJoinRel(ImplicitLocOpBuilder builder, const Rel &message) {
   const HashJoinRel &hashJoinRel = message.hash_join();
-
-  // Import left and right inputs.
   const Rel &leftRel = hashJoinRel.left();
   const Rel &rightRel = hashJoinRel.right();
 
@@ -686,13 +684,11 @@ importHashJoinRel(ImplicitLocOpBuilder builder, const Rel &message) {
   if (failed(leftOp) || failed(rightOp))
     return failure();
 
-  // Build `HashJoinOp`.
   Value leftVal = leftOp.value()->getResult(0);
   Value rightVal = rightOp.value()->getResult(0);
 
   std::optional<JoinType> join_type = static_cast<JoinType>(hashJoinRel.type());
 
-  // Check for unsupported set operations.
   if (!join_type)
     return mlir::emitError(builder.getLoc(), "unexpected 'operation' found");
 


include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -716,6 +717,109 @@ def Substrait_FilterOp : Substrait_RelOp<"filter", [
...
+    %4 = field_reference %arg1[0] : tuple<si32, si32, si32>
+    %5 = compare not_distinct_from %3, %4 : (si32, si32) -> si1
+    ```
+  }];
+  
+  let arguments = (ins
+    Substrait_ExpressionType:$lhs,
+    Substrait_ExpressionType:$rhs,
+    OptionalAttr<SimpleComparisonType>:$comparison_type,
+    OptionalAttr<UI32Attr>:$custom_function_id
```
mortbopet, line 741: If `custom_function_id` is deliberately not implemented, i'd say remove this argument and write a TODO in the `description` field.

include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -716,6 +717,109 @@ def Substrait_FilterOp : Substrait_RelOp<"filter", [
...
+  );
+  
+  let results = (outs SI1:$result);
+  
+  let assemblyFormat = [{
+    $comparison_type $lhs `,` $rhs attr-dict `:` `(` type($lhs) `,` type($rhs) `)` `->` type($result)
```
mortbopet, line 747:
```
`(` type($lhs) `,` type($rhs) `)` `->` type($result)
```
could be
```
functional-type(operands, $result)

File: include/substrait-mlir/Dialect/Substrait/IR/SubstraitOps.td [hunk 3]
```diff
@@ -716,6 +717,109 @@ def Substrait_FilterOp : Substrait_RelOp<"filter", [
...
+    JoinType:$join_type,
+    OptionalAttr<Substrait_AdvancedExtensionAttr>:$advanced_extension
+  );
+  let regions = (region AnyRegion:$condition);
+  let hasRegionVerifier = 1;
+  let results = (outs Substrait_Relation:$result);
+  let assemblyFormat = [{
+    $join_type $left `,` $right
+    (`advanced_extension` `` $advanced_extension^)?
+    attr-dict `:` type($left) `,` type($right) `->` type($result) $condition
```
mortbopet, line 807: as above (functional-type).

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -1035,6 +1035,107 @@ LiteralOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+
+  auto resultType = TupleType::get(context, fieldTypes);
+
+  inferredReturnTypes = SmallVector<Type>{resultType};
+
+  return success();
+}
+
+LogicalResult KeyComparisonOp::verify() {
+  auto &res =
```
mortbopet, line 1081: Since this may be a bit unintuitive at first (i.e. one might expect that the comparison op should always have identical types for both operands), a comment might be warranted here.

Do we have some documentation about substraits stance here? That is, this code is assuming that the comparison op is able to perform comparisons against types that are "cast-compatible" (int <=> decimal, string <=> varchar).

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -1035,6 +1035,107 @@ LiteralOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+            if (!mlir::isa<IntegerType, substrait::DecimalType>(getRhs()))
+              return this->emitError("Invalid rhs type for integer comparison, "
+                                     "expected integer-like type but got")
+                     << getRhs().getType();
+            return success();
+          });
+  if (failed(res))
+    return res;
+
+  return success();
```
mortbopet, line 1101: Could just `return res;` here.

lib/Dialect/Substrait/IR/Substrait.cpp
mortbopet, line 1081: Nit: Storing `LogicalResult` by reference is atypical; i'd just do `auto res` here.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -1035,6 +1035,107 @@ LiteralOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+}
+
+LogicalResult HashJoinOp::verifyRegions() {
+  if (getCondition().empty())
+    return emitOpError() << "hash join must have a condition region";
+
+  // Verify that we have exactly two arguments in the region, matching the left
+  // and right relations
+  Block &block = getCondition().front();
```
mortbopet, line 1110: Since you've added the `SingleBlockImplicitTerminator` trait, you should be able to get the body block by just saying `Block* body = getBody()`.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -1035,6 +1035,107 @@ LiteralOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+  Type yieldedType = yieldOp.getValue()[0].getType();
+  MLIRContext *context = getContext();
+  Type si1Type = IntegerType::get(context, 1, IntegerType::Signed);
+  if (yieldedType != si1Type)
+    return emitOpError() << "condition region must yield a boolean (si1) value";
+
+  Value conditionValue = yieldOp.getValue()[0];
+  if (!isa<KeyComparisonOp>(conditionValue.getDefiningOp())) {
+    return emitOpError() << "condition must be produced by a comparison";
+  }
```
mortbopet, line 1134: Nit: remove braces.

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -757,6 +758,86 @@ FailureOr<std::unique_ptr<Rel>> SubstraitExporter::exportOperation(JoinOp op) {
...
+
+  auto hashJoinRel = std::make_unique<HashJoinRel>();
+  hashJoinRel->set_allocated_common(relCommon.release());
+  hashJoinRel->set_allocated_left(leftRel->release());
+  hashJoinRel->set_allocated_right(rightRel->release());
+  hashJoinRel->set_type(static_cast<HashJoinRel::JoinType>(op.getJoinType()));
+
+  if (op.getCondition().empty()) {
+    return op->emitOpError("missing join condition");
+  }
```
mortbopet, line 795: nit: remove braces

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -757,6 +758,86 @@ FailureOr<std::unique_ptr<Rel>> SubstraitExporter::exportOperation(JoinOp op) {
...
+
+  Block &conditionBlock = op.getCondition().front();
```
mortbopet, line 797: As above (`op.getBody()`).

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -757,6 +758,86 @@ FailureOr<std::unique_ptr<Rel>> SubstraitExporter::exportOperation(JoinOp op) {
...
+  Operation *terminator = conditionBlock.getTerminator();
+  Value conditionValue = cast<YieldOp>(terminator).getValue()[0];
+
+  auto compareOp =
+      dyn_cast_or_null<KeyComparisonOp>(conditionValue.getDefiningOp());
+  if (!compareOp) {
+    return op->emitOpError("join condition must be a KeyComparisonOp");
+  }
```
mortbopet, line 805: This could be removed, since you're already checking for this in your verifier. 
nit: remove braces.

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -757,6 +758,86 @@ FailureOr<std::unique_ptr<Rel>> SubstraitExporter::exportOperation(JoinOp op) {
...
+    return op->emitOpError() << "left key must be a field reference";
+  }
+
+  FailureOr<std::unique_ptr<Expression>> rightKeyExpr;
+  if (auto rightFieldRef =
+          dyn_cast_or_null<FieldReferenceOp>(rightKey.getDefiningOp())) {
+    rightKeyExpr = exportOperation(rightFieldRef);
+  } else {
+    return op->emitOpError() << "right key must be a field reference";
+  }
```
mortbopet, line 823: This logic should be moved to a verifier of the `KeyComparisonOp` (i.e. set `hasVerifier = 1` for the op).

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -633,6 +642,115 @@ static mlir::FailureOr<JoinOp> importJoinRel(ImplicitLocOpBuilder builder,
...
+
+  if (!join_type)
+    return mlir::emitError(builder.getLoc(), "unexpected 'operation' found");
+
+  mlir::FailureOr<HashJoinOp> hashJoinOp =
+      builder.create<HashJoinOp>(leftVal, rightVal, *join_type);
+
+  if (failed(hashJoinOp)) {
+    return failure();
+  }
```
mortbopet, line 700: as above

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -633,6 +642,115 @@ static mlir::FailureOr<JoinOp> importJoinRel(ImplicitLocOpBuilder builder,
...
+
+    Value leftKey = leftKeyOp.value();
+    Value rightKey = rightKeyOp.value();
+
+    condition = builder.create<KeyComparisonOp>(leftKey, rightKey,
+                                                simpleComparisonType);
+  } else {
+    // TODO(trion): Handle custom function if present
+    return mlir::emitError(builder.getLoc(),
+                           "custom comparison functions not yet supported");
```
mortbopet, line 734: Nit: change to `"custom comparison functions for hash_join not yet supported"` to make it a bit more clear, when one has a very large substrait input file, and it may be a bit hard to decipher what the error message refers to.

mortbopet (commented):
Very nice work! left a bunch of comments. Obviously there needs to be import/export/IR tests, but given the draft state, i assume those are in the pipeline :).

The post-join filter is definitely a funny thing here, and principally i agree with your approach (creating a filter node after the hash-join). It's a normalization of the IR, and it avoids introducing redundant operations (i.e. if we'd actually create a post-join filter in the hash-join op).
However, it does present an issue for round-tripping through the IR, which this would break. It's obviously not impossible to fix this (one way would be to add extra logic in the export pass to detect how a filter op is being used, and infer the post-join filter from there)... @ingomueller-net thoughts?

PR #131, opened 2025-05-16
feat: implement short-hand form for types

ingomueller-net:
This PR adds a short-hand version for the assymbly format of the types in the Substrait dialect and adapts the tests accordingly.  The implementation follows the same pattern as the one in the [LLVMIR dialect](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/LLVMIR/IR/LLVMTypeSyntax.cpp) upstream. With that format, we can write `interval_ds` instead of `!substrait.interval_day_second`, for example. The short-hand versions mainly consist of just the op name e.g., `date` instead of `!substrait.date` but a few short-hand versions are shortend further (like `interval_ds`). The short-hand versions can be used everywhere were the custom `SubstraitType` printer and parser is used. For now, this includes all `RelOp`s, the `YieldOp`, as well as types tested inside `RelationType`, which covers all occurrences of types in the test suite except for the types of `StringAttr`s, such as in `literal "foo" : !substrait.string`.  This PR does also not change the assembly format of attributes, which have a similar source of redundancy that we may want to remove in follow-up work.

### DIFF
--- a/test/Target/SubstraitPB/Import/literal.textpb
+++ b/test/Target/SubstraitPB/Import/literal.textpb
@@ -13,12 +13,12 @@
 # CHECK:   substrait.plan version 0 : 42 : 1 {
 # CHECK-NEXT:           relation {
 # CHECK-NEXT:             %[[VAL_0:.*]] = named_table @t1 as ["a"] : rel<f32>
-# CHECK-NEXT:             %[[VAL_1:.*]] = project %[[VAL_0]] : rel<f32> -> rel<f32, !substrait.decimal<3, 2>> {
+# CHECK-NEXT:             %[[VAL_1:.*]] = project %[[VAL_0]] : rel<f32> -> rel<f32, decimal<3, 2>> {
 # CHECK-NEXT:             ^bb0(%[[VAL_2:.*]]: tuple<f32>):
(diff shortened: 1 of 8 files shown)


test/Target/SubstraitPB/Import/literal.textpb
```diff
@@ -463,13 +463,13 @@ version {
 # CHECK:      substrait.plan version 0 : 42 : 1 {
 # CHECK-NEXT:   relation
 # CHECK:         %[[V0:.*]] = named_table
-# CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : rel<si1> -> rel<si1, !substrait.timestamp, !substrait.timestamp_tz> {
+# CHECK-NEXT:    %[[V1:.*]] = project %[[V0]] : rel<si1> -> rel<si1, timestamp, timestamp_tz> {
 # CHECK-NEXT:    ^[[BB0:.*]](%[[ARG0:.*]]: tuple<si1>):
 # CHECK-NEXT:      %[[V2:.*]] = literal #substrait.timestamp<10000000000us>
```
dshaaban01, line 468: do the `#substrait.TYPE` not get shortened to `TYPE`?

dshaaban01 (approved):
This is great!

PR #140, opened 2025-05-27
build: set up clang-tidy

ingomueller-net:
This PR sets up clang-tidy to be run by CMake on our targets. Currently
all checks that fail are disabled such that CI continues to pass.


ingomueller-net:
Yeah, right? And it even has the "apply fixes" workflow integrated, which will make the pending improvements really easy, I think.

PR #143, opened 2025-06-02
style: enable and reconfigure`readability-braces-around-statements`

ingomueller-net:
That flag was disabled when `clang-tidy` was set up initially in #140 because it failed in too many places. This PR enables it, such that identifier naming will be enforced in the future. It also changes the size of "short" statements, which don't need braces, to three lines because all examples of two lines that the tool wanted to change seemed perfectly readable as is to me. Finally, the PR applies the automatic fixes such that the code is compliant with the new configuration.


.clang-tidy
```diff
@@ -1,5 +1,6 @@
 ---
 # Enable the same checks as MLIR, except for the following:
+#    misc-include-cleaner
 #    readability-braces-around-statements
```
dshaaban01, line 4: Should this line be deleted since the flag is now enabled through this PR?

dshaaban01 (approved):

.clang-tidy
ingomueller-net: Good catch. Done during last rebase.

PR #146, opened 2025-06-02
build: enable `misc-use-anonymous-namespace` check of `clang-tidy`

ingomueller-net:
That flag was disabled when `clang-tidy` was set up initially in #140 because it failed in too many places. This PR enables it and fixes all instances found by `clang-tidy` such that the style will be enforced in the future.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,7 +1,6 @@
 ---
 # Enable the same checks as MLIR, except for the following:
 #    readability-braces-around-statements
-#    misc-use-anonymous-namespace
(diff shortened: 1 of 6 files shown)


.clang-tidy
```diff
@@ -1,7 +1,6 @@
 ---
 # Enable the same checks as MLIR, except for the following:
 #    readability-braces-around-statements
-#    misc-use-anonymous-namespace
 # TODO(ingomueller): fix issues found by clang-tidy and enable checks.
```
dshaaban01, line 4: Change/ alter this TODO maybe?

PR #153, opened 2025-06-16
ci(clang-format): handle system headers with `-` and `_` correctly

ingomueller-net:
The regex for defining the ordering for system headers did not contain the `-` and `_` characters to `#include <type_traits>` was not ordered correctly. This PR fixes that.


lib/Target/SubstraitPB/ProtobufUtils.h
```diff
@@ -9,10 +9,10 @@
 #ifndef LIB_TARGET_SUBSTRAITPB_PROTOBUFUTILS_H
 #define LIB_TARGET_SUBSTRAITPB_PROTOBUFUTILS_H
 
-#include <type_traits>
-
 #include "mlir/IR/Location.h"
 
+#include <type_traits>
+
```
dshaaban01, line 15: Why this change? (I know it is harmless - but am just curious).

dshaaban01 (approved):

Issue #158, opened 2025-07-01
Set up CI without assertions and fix warnings about unused variables


ingomueller-net:
Some variables are only used in assertions, so if we build without assertions, we get warnings about them being unused, and they turn into errors if we compile with `-Werror`. We should set up a (nightly?) CI target that builds the project without assertions to catch and remove these.

PR #160, opened 2025-07-05
ci: set up build without assertions as post-merge CI job

ingomueller-net:
This PR adds another CI target that compiles the project without assertions and that is only enabled as a post-merge check. This addresses substrait-io#158, which tracks the fact that some warnings are only raised when a project is compiled without assertions, most notably when variables are only used by assertions and then become unused. This is a rare and minor issue, so the delay that would be caused by checking for every commit of every PR would not be worth it. The post-merge check is achieved by turning the main CI job for compiling into a `matrix` whose value sets the flag that enables assertions in CMake and then excluding the `ON` entry for all but `push` events (which are triggered after a PR is "pushed" onto main during merge).

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -13,6 +13,14 @@ jobs:
   ubuntu-build-and-test:
     name: Ubuntu build and test (Release Asserts)
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        # Build with and without asserts but exclude the latter for PRs.
+        llvm_enable_assertions:
+          - "ON"
+          - "OFF"
+        exclude:
+          - llvm_enable_assertions: ${{ github.event_name != 'push' && 'OFF' }}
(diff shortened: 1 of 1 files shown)


ingomueller-net:
It' s difficult to test whether this actually triggers after merge. However, I have tested that the condition depending on the event type works by changing it to `exclude: - cmake_build_type: ${{ github.event_name != 'pull_request' && 'Release' }}`. This did *not* exclude the `Release` build in the `pull_request` event (which is triggered every time a PR is updated), so this [CI job](https://github.com/substrait-io/substrait-mlir-contrib/actions/runs/16088670401/job/45402873751?pr=160) did execute the corresponding job.

ingomueller-net:
Seems like it worked 😄 The post-merge job [found](https://github.com/substrait-io/substrait-mlir-contrib/actions/runs/16303093941/job/46042421591) the problem mentioned in #158.

PR #173, opened 2025-09-05
ci(dependabot): update all packages weekly

ingomueller-net:
This PR changes the configuration of dependabot to update all packages weekly. I originally wanted to update only LLVM weekly and everything else daily but I coudln't find out the right way to do this, so I am giving up. I rather only update all packages weekly than having to update LLVM every single day.


dshaaban01 (approved):
👍

Issue #169, opened 2025-08-19
Use `static` for file-private functions instead of anonymous namespaces


ingomueller-net:
As per the [LLVM Coding Standards](https://llvm.org/docs/CodingStandards.html#restrict-visibility), we should use the `static` keyword to restrict the visibility of functions to the current file instead of putting them into an anonymous namespace. Currently, `clang-tidy` enforces the latter approach, so we should change its configuration and apply the LLVM-style solution.

PR #180, opened 2025-12-16
fix: `llvm-prefer-static-over-anonymous-namespace` [WIP]

dshaaban01:
Use the static keyword to restrict the visibility of functions to the current file instead of putting them into an anonymous namespace. Update `.clang-tidy` to `llvm-prefer-static-over-anonymous-namespace`.

This PR makes this change for all .cpp files (EXCEPT import.cpp). 
Addresses[ issue 169](https://github.com/substrait-io/substrait-mlir-contrib/issues/169).

### DIFF
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -177,14 +177,12 @@ LogicalResult mlir::substrait::DecimalType::verify(
   return success();
 }
 
-namespace {
 // Count the number of digits in an APInt in base 10.
-size_t countDigits(const APInt &value) {
+static size_t countDigits(const APInt &value) {
   llvm::SmallVector<char> buffer;
   value.toString(buffer, 10, /*isSigned=*/false);
   return buffer.size();
 }
-} // namespace
 
 LogicalResult mlir::substrait::DecimalAttr::verify(
     llvm::function_ref<mlir::InFlightDiagnostic()> emitError, DecimalType type,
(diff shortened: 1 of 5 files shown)


lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -719,30 +721,24 @@ void printSubstraitType(AsmPrinter &printer, Operation *op,
...
 
-namespace {
-
 /// Computes the type of the nested field of the given `type` identified by
 /// `position`. Each entry `n` in the given index array `position` corresponds
 /// to the `n`-th entry in that level.
-FailureOr<Type> computeTypeAtPosition(Location loc, Type type,
-                                      ArrayRef<int64_t> position);
-
-namespace impl {
```
ingomueller-net: We should keep this to make clear that the functions in this namespace are an implementation detail.

lib/Target/SubstraitPB/ProtobufUtils.cpp
```diff
@@ -27,12 +27,10 @@ namespace mlir::substrait::protobuf_utils {
 
 namespace pb = ::google::protobuf;
 
-namespace {
 template <typename RelType>
 static const RelCommon *getCommon(const RelType &rel) {
```
ingomueller-net: *If* we decide to keep the anonymous namespaces, we should probably remove the `static` keyword here...

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -168,13 +170,15 @@ void SubstraitExporter::exportAdvancedExtension(ExtensibleOpInterface op,
 
   // Set `optimization` field if present.
   if (optimizationAttr) {
-    std::unique_ptr<pb::Any> optimization = exportAny(optimizationAttr);
+    std::unique_ptr<::google::protobuf::Any> optimization =
```
ingomueller-net: Hmm... this is another disadvantage of not using anonymous namespaces, which I didn't see mentioned in the resources I have linked to. Not being able to use namespace alias does make some code more verbose. Putting the alias outside of the namespace also has it problems (though I don't remember exactly what it was).

